PR #25, opened 2021-09-20
Make username and password in SMTP driver optional

sfc-gh-pkommini:
Make username and password optional

Tests:

Tested HTTP Driver:
image

Tested SMTP Driver:
image

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -73,12 +73,12 @@ def process_row(
 
         # We make unauthenticated request if the 'host' key is missing.
         if not auth_host:
-            LOG.info(f"'auth' missing the \'host\' key.")
+            LOG.info(f"'auth' missing the 'host' key.")
 
         # We reject the request if the 'auth' doesn't match the pinned host.
         if auth_host != req_host:
             raise ValueError(
-                'Requests can only be made to host provided in the auth header.'
+                "Requests can only be made to host provided in the auth header."
             )
 
         elif 'basic' in req_auth:
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -73,7 +73,7 @@ def process_row(
         auth_host = req_auth.get('host')
 
         if not auth_host:
-            raise ValueError('\'auth\' is missing the \'SMTP_HOST\' key.')
+            raise ValueError("'auth' is missing the 'host' key.")
 
         # We reject the request if the 'auth' is not pinned to a host or doesn't match the pinned host
         if auth_host != req_host:
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -73,7 +73,7 @@ def process_row(
 
         # We make unauthenticated request if the 'host' key is missing.
         if not auth_host:
-            LOG.info(f'\'auth\' missing the \'host\' key.')
+            LOG.info(f"'auth' missing the \'host\' key.")
 
         # We reject the request if the 'auth' doesn't match the pinned host.
         if auth_host != req_host:
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -59,6 +59,7 @@ def process_row(
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
 
+    # We look for an auth header and if found we parse it from it's encoded format
     if auth is not None:
         auth = decrypt_if_encrypted(auth)
         assert auth is not None
@@ -73,6 +74,7 @@ def process_row(
         if 'host' in req_auth and not req_host:
             req_host = req_auth['host']
 
+        # We reject the auth header if not pinned to a host
         if req_auth.get('host') != req_host:
             pass  # if host in ct, only send creds to that host
         elif 'basic' in req_auth:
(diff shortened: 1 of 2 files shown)


lambda_src/drivers/process_smtp.py
```diff
@@ -1,16 +1,72 @@
...
 import smtplib
 import ssl
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
+from typing import Dict, Tuple, Any
 
 from ..vault import decrypt_if_encrypted
 
 
+def parse_smtp_creds(auth_dict: Dict) -> Tuple[Any, Any, Any]:
```
sfc-gh-afedorov, line 11: could we be a little more prescriptive and also make sure we can't possibly confuse some other drivers' credentials to be re-used here? maybe e.g. add a `"driver": "smtp"` key or just force `smtp_host` to make sure another driver wouldn't re-use these credentials?

sooner or later we'll need to document this and having a single option would make things simpler there, as well. people shouldn't be guessing here, and docs should give them one right answer.

that said, do appreciate the cool next/map combo, of course :)

sfc-gh-afedorov (approved):
lgtm, minor feedback re: fewer options being better

lambda_src/drivers/process_https.py
```diff
@@ -59,6 +58,7 @@ def process_row(
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
 
+    # We look for an auth header and if found we parse it from it's encoded format
```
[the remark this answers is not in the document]
sfc-gh-pkommini: ACK.

lambda_src/drivers/process_smtp.py
```diff
@@ -27,6 +37,15 @@ def process_row(
     password = decrypt_if_encrypted(password)
     sender_email = sender_email or user
 
+    if auth:
+        auth_json = decrypt_if_encrypted(auth)
+        auth_dict = json.loads(auth_json)
+        user, password, host = parse_smtp_creds(auth_dict)
```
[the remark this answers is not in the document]
sfc-gh-pkommini: There is nothing to match the host with. As there is no host passed to the driver from the external function, so the threat model here is a bit different in that if the `auth` dict doesn't have an `SMTP_HOST` we reject the request.

lambda_src/drivers/process_smtp.py
```diff
@@ -27,6 +37,15 @@ def process_row(
...
+        if not host:
+            raise ValueError(
+                'Credentials not pinned to an SMTP host! This can increase the risk of credential compromise.'
```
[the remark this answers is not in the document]
sfc-gh-pkommini: ACK.

lambda_src/drivers/process_https.py
```diff
@@ -59,22 +58,29 @@ def process_row(
...
             loads(auth)
             if auth.startswith('{')
             else parse_header_dict(auth)
             if auth
             else {}
         )
+        auth_host = req_auth.get('host')
+
+        if not auth_host:
+            raise ValueError('\'auth\' is missing the \'SMTP_HOST\' key.')
```
sfc-gh-afedorov, line 76: - use `"` for strings that are shown to users
- the key is `'host'` here

lambda_src/drivers/process_https.py
```diff
@@ -59,22 +58,29 @@ def process_row(
...
 
-        if 'host' in req_auth and not req_host:
-            req_host = req_auth['host']
+        # We reject the request if the 'auth' is not pinned to a host or doesn't match the pinned host
+        if auth_host != req_host:
```
sfc-gh-afedorov, line 79: keep `auth_host is Null` in this check?

sfc-gh-afedorov: that's the opposite of what we want -- if the auth host is missing, we want to send an un-authenticated https request

sfc-gh-afedorov: in general, can we keep refactoring of https in a separate PR? this one is supposed to be a change to the SMTP handler

lambda_src/drivers/process_smtp.py
```diff
@@ -75,13 +94,17 @@ def process_row(
 
     try:
         result = smtpserver.sendmail(sender_email, recipients, message.as_string())
+    except ValueError as e:
```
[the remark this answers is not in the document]
sfc-gh-afedorov: could we put just that part into a try/except, then, and have a separate try for sendmail?

lambda_src/drivers/process_smtp.py
```diff
@@ -27,6 +39,20 @@ def process_row(
     password = decrypt_if_encrypted(password)
     sender_email = sender_email or user
 
+    if auth:
+        auth_json = decrypt_if_encrypted(auth)
+        assert auth is not None
```
sfc-gh-afedorov, line 44: when would this be the case? iirc decrypt raises an exception if it cannot decrypt, and returns `None` only if `auth` is `None`

lambda_src/drivers/process_https.py
sfc-gh-pkommini: ACK. I was making the pinning uniform for both. I thought this PR was related to our pinning discussion hence the changes related to that.

lambda_src/drivers/process_smtp.py
sfc-gh-pkommini: I tried it make it uniform with the HTTPS driver but yeah will remove the assertion in both if that's the case.

lambda_src/drivers/process_https.py
sfc-gh-afedorov: thanks, yes -- we want user / password to be optional in SMTP. we want to retain the optional auth in https

sfc-gh-pkommini: It's done. Let me know if the changes look okay. Opened a new PR for the HTTP changes.

sfc-gh-pkommini: Tested with secrets manager. 
image

PR #28, opened 2021-10-14
Refactors to https driver

sfc-gh-pkommini:
- Uniform driver pinning interface and error messages
- Uniform `None` checking with SMTP driver

**_Tests_**:
Success test (if `auth` dict has `host` key to pin the host to which we can send auth credentials):
image

Failure test (if `auth` dict doesn't have the `host` key):
image

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -33,7 +33,6 @@ def process_row(
     verbose: bool = False,
     cursor: Text = '',
     results_path: Text = '',
-    destination_uri: Text = '',
 ):
     if not base_url and not url:
         raise ValueError('Missing required parameter. Need one of url or base-url.')
@@ -59,22 +58,29 @@ def process_row(
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
 
+    # We look for an auth header and if found, we parse it from its encoded format
     if auth is not None:
         auth = decrypt_if_encrypted(auth)
-        assert auth is not None
+
         req_auth = (
             loads(auth)
-            if auth.startswith('{')
+            if auth and auth.startswith('{')
             else parse_header_dict(auth)
             if auth
             else {}
         )
+        auth_host = req_auth.get('host')
 
-        if 'host' in req_auth and not req_host:
-            req_host = req_auth['host']
+        # We make unauthenticated request if the 'host' key is missing.
+        if not auth_host:
+            LOG.info(f"'auth' missing the 'host' key.")
 
-        if req_auth.get('host') != req_host:
-            pass  # if host in ct, only send creds to that host
+        # We reject the request if the 'auth' doesn't match the pinned host.
+        if auth_host != req_host:
+            raise ValueError(
+                "Requests can only be made to host provided in the auth header."
+            )
+
         elif 'basic' in req_auth:
             req_headers['Authorization'] = make_basic_header(req_auth['basic'])
         elif 'bearer' in req_auth:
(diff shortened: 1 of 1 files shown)


sfc-gh-pkommini:
@sfc-gh-afedorov 

Could you describe what is the behavior you expect in the HTTPS driver when `auth_host is None`. I would think auth failure is not what you expect when you have the right credentials but have the wrong `host` key.
image

It seems to error out with invalid credentials as opposed to we're failing your authentication because you don't have a `host` key passed in. This seems like it will confuse users to thinking they have wrong credentials.

lambda_src/drivers/process_https.py
```diff
@@ -33,7 +33,6 @@ def process_row(
     verbose: bool = False,
     cursor: Text = '',
     results_path: Text = '',
-    destination_uri: Text = '',
```
sfc-gh-afedorov, line 36: why remove the destination URI?

lambda_src/drivers/process_https.py
```diff
@@ -59,22 +58,28 @@ def process_row(
...
 
+    # We look for an auth header and if found, we parse it from its encoded format
     if auth is not None:
         auth = decrypt_if_encrypted(auth)
-        assert auth is not None
+
         req_auth = (
             loads(auth)
-            if auth.startswith('{')
+            if auth and auth.startswith('{')
```
sfc-gh-afedorov, line 67: if the assert helps the type system know this is Ok without explicitly stating it that seems better -- maybe add a comment that this is why it's there?

lambda_src/drivers/process_https.py
```diff
@@ -59,22 +58,28 @@ def process_row(
...
-        if req_auth.get('host') != req_host:
-            pass  # if host in ct, only send creds to that host
+        # We reject the request if the 'auth' is present but doesn't match the pinned host.
+        if auth_host and auth_host != req_host:
+            raise ValueError(
+                "Requests can only be made to host provided in the auth header."
+            )
+        # We make unauthenticated request if the 'host' key is missing.
+        elif not auth_host:
+            LOG.info(f"'auth' missing the 'host' key.")
```
sfc-gh-afedorov, line 81: these logs go to CloudWatch, which isn't going to get sent to the user. agreed with sentiment that the silent failure can be bad UX, so maybe returning the error that auth credentials must pin a host is better?

lambda_src/drivers/process_https.py
sfc-gh-pkommini: Sorry I don't think I fully understood the requested change here. Just to confirm, you're asking to bring back the assert and remove the `None` check and add a comment as to why the assert is there? 

assert fail causes an error and `if auth` check only checks for a curly branch when auth is not `None`. Isn't that the behavior we want here?

sfc-gh-afedorov: yeah -- I think the extra check is vestigial since we know it's not going to `String` and the `assert` was there to make sure the pypi type system doesn't complain since it doesn't have a way of indicating that the decrypt will return None only if it's passed None.

lambda_src/drivers/process_https.py
```diff
@@ -59,22 +58,27 @@ def process_row(
...
-            req_host = req_auth['host']
-
-        if req_auth.get('host') != req_host:
-            pass  # if host in ct, only send creds to that host
+        # We reject the request if the 'auth' is present but doesn't match the pinned host.
+        if auth_host and auth_host != req_host:
+            raise ValueError(
+                "Requests can only be made to host provided in the auth header."
+            )
+        # We make unauthenticated request if the 'host' key is missing.
```
sfc-gh-afedorov, line 79: comment now out of date

sfc-gh-afedorov (approved):

PR #31, opened 2021-11-03
Add dockerfile to build geff image to run in lambda docker

sfc-gh-pkommini:
docker file to build geff as a docker image:

_**Testing**_:
abuse IP DB scan works:
image

ecr list_repos works:
image

With storage pipeline works:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM public.ecr.aws/lambda/python:3.8
+
+# Install git and git clone the geff repo
+RUN yum update -y && \
+    yum install -y git && \
+    rm -Rf /var/cache/yum
+
+COPY lambda_src/ "${LAMBDA_TASK_ROOT}"
+
+COPY requirements.txt "${LAMBDA_TASK_ROOT}"
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM public.ecr.aws/lambda/python:3.8
```
sfc-gh-afedorov, line 1: would 3.10 be ok?

sfc-gh-afedorov (approved):

Dockerfile
sfc-gh-pkommini: Yeah I don't see any issues. Will test and push the change.

sfc-gh-pkommini: Latest seems to be python3.9: https://docs.aws.amazon.com/lambda/latest/dg/python-image.html

sfc-gh-pkommini: And it works just fine.

sfc-gh-afedorov: Got it -- is there no public spec w.r.t. bases and do they all have to run using public.ecr.aws? py310 was released about a month ago, so it seems they have not yet gotten to releasing the lambda base for it.

PR #40, opened 2022-01-18
Add assume role option to boto3 driver

sfc-gh-afedorov:
tested against our local and dev account [here](https://app.snowflake.com/us-west-2/snowhouse/w3RZuCaZvmW3)

### DIFF
--- a/lambda_src/drivers/process_boto3.py
+++ b/lambda_src/drivers/process_boto3.py
@@ -3,7 +3,7 @@ from json import loads
 import boto3
 from botocore.response import StreamingBody
 
-from utils import pick
+from ..utils import pick
 
 DISALLOWED_CLIENTS = {'kms', 'secretsmanager'}
 
@@ -35,7 +35,7 @@ def process_row(
                 RoleSessionName='geff'
                 if role_session_name is None
                 else f'geff_{role_session_name}',
-                **assume_role_params
+                **assume_role_params,
             )
         client = boto3.client(
             client_name,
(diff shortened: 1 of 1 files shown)


lambda_src/drivers/process_boto3.py
```diff
@@ -3,21 +3,53 @@
...
+from utils import pick
 
 DISALLOWED_CLIENTS = {'kms', 'secretsmanager'}
 
 
 def process_row(
-    client_name, method_name, results_path=None, region='us-west-2', **kwargs
+    client_name,
+    method_name,
+    assume_role_arns=None,
```
sfc-gh-pkommini, line 14: Is this always ARNs? Is it always a list? What possible values can this variable take?

If it is a sequence of roles that need to be assumed would it make sense to add the keyword `sequence` or `chain` in the variable to indicate that?

sfc-gh-pkommini (approved):
lgtm but for cosmetic variable name clarity.

lambda_src/drivers/process_boto3.py
```diff
@@ -3,21 +3,53 @@
 import boto3
 from botocore.response import StreamingBody
 
+from utils import pick
```
sfc-gh-pkommini, line 6: This should be a relative import.

sfc-gh-afedorov: good catch -- was testing without top level package

lambda_src/drivers/process_boto3.py
sfc-gh-afedorov: it's a sequence of roles, yes, with either strings or objects which are going to be passed into `assume_role`, e.g. `{"RoleArn":"...","ExternalId":"..."}`

sfc-gh-afedorov:
tested also using `make build`

PR #46, opened 2022-05-11
Create the console_logger, geff_logger, sentry_driver_logger 

sfc-gh-pkommini:
Create Loggers and then use it to send events from appropriate places.


lambda_src/drivers/process_google.py
```diff
@@ -1,7 +1,7 @@
 from json import loads
 
-from google.oauth2 import credentials, service_account  # type: ignore
-from googleapiclient.discovery import build  # type: ignore
```
sfc-gh-afedorov, line 4: why remove ignoring of type? is there a new type definition somewhere?

sfc-gh-pkommini: No. Sorry. Was trying to see if latest version of the stubs has typing. It still doesn't.

lambda_src/lambda_function.py
```diff
@@ -186,22 +193,26 @@ def lambda_handler(event: Any, context: Any) -> Dict[Text, Any]:
...
+        if method == 'POST' and destination:
+            return async_flow_init(event, context)
+        elif method == 'POST':
+            return sync_flow(event, context)
+        elif method == 'GET':
+            return async_flow_poll(destination, batch_id)
+
+        return create_response(400, 'Unexpected Request.')
+    except Exception as e:
+        GEFF_SENTRY_LOGGER.exception(e)
```
sfc-gh-afedorov, line 217: what if SENTRY isn't set up in a deployment. not every user will use it.

sfc-gh-afedorov (commented):
could you split this up into separate PR's with clear singular objectives? specifically, could we make a PR for just the driver for GSE-2167? I can see there being value in also making an optional feature to send un-handled stack traces to Sentry, but that's a different ask and not something we've run into trouble with in production much, since driver errors are reported back to the user.

sfc-gh-pkommini:
> could you split this up into separate PR's with clear singular objectives? specifically, could we make a PR for just the driver for GSE-2167? I can see there being value in also making an optional feature to send un-handled stack traces to Sentry, but that's a different ask and not something we've run into trouble with in production much, since driver errors are reported back to the user.

@sfc-gh-afedorov  What specific change are you referring to split into a separate PR?

lambda_src/lambda_function.py
sfc-gh-pkommini: The logger is setup regardless of Sentry setup. Specific loggers are channeled into specific projects in Sentry. The `sentry_sdk` is initialized only if `SENTRY_DSN` and `GEFF_DSN` env variables are set.

sfc-gh-pkommini: [Here](https://github.com/Snowflake-Labs/geff/blob/9b1c9b77730a10a63b66aac2540566a1f855e7b0/lambda_src/utils.py#L129-L159) is the definition of the `setup_sentry`.

lambda_src/drivers/process_https.py
```diff
@@ -168,19 +171,22 @@ def process_row(
                     else response_body
                 ),
             }
+            GEFF_SENTRY_LOGGER.exception(e)
```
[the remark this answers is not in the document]
sfc-gh-pkommini: Makes sense. 400s are definitely something we want the user to see. 500s and unhandled 400s are what we need to be sent to `geff` project on Sentry.

sfc-gh-pkommini:
@sfc-gh-afedorov The reason why we're `send_event` to send to Sentry based on the logger is because we need to send to 2 different DSNs for the same project which is not an intended use case.

However, [this discussion ](https://github.com/getsentry/sentry-python/issues/198) captures a similar use case. Let me know if that helps clarify why the logging changes and Sentry driver are tied into the same PR.

lambda_src/drivers/process_sentry_logger.py
```diff
@@ -77,16 +77,17 @@ def process_row(
...
         schema (str): The schema.
         name (str): The object name.
-        ts (str): The timestamp of the error.
         history_type (str): pipe or task or query.
-        region (str): Region of the Snowflake account.
-        account (str): The Snowflake account name.
+        error (str): The actual error message.
+        ts (str): The timestamp of the error.
     """
+    CONSOLE_LOGGER.info(f'sentry_logger driver invoked.')
```
sfc-gh-afedorov, line 89: what's the logging strategy here? is this something we want to log on every driver? if so, can we put it into the invoking code, and make it debug rather than info?

lambda_src/drivers/process_sentry_logger.py
```diff
@@ -77,16 +77,17 @@ def process_row(
...
+
     history_url = get_snowsight_url(
```
sfc-gh-afedorov, line 91: I don't think we should have Snowsight hard-coded into the Sentry driver -- can we move this code to the Snowflake UDF?

lambda_src/drivers/process_sentry_logger.py
```diff
@@ -113,5 +116,8 @@ def process_row(
             sentry_sdk.set_tag('error_time', ts)
             sentry_sdk.set_tag('history_type', history_type)
             SENTRY_DRIVER_LOGGER.exception(error)
+        status = f'Captured {name} error at {ts}.'
     except Exception as e:
```
sfc-gh-afedorov, line 120: this is a little confusing -- which exceptions are you looking to capture here, exactly?

lambda_src/drivers/process_sentry_logger.py
```diff
@@ -95,6 +96,8 @@ def process_row(
         name,
         history_type
     )
+    CONSOLE_LOGGER.info(f'Constructed history_url: {history_url}.')
+    status: Optional[str] = None
```
sfc-gh-afedorov, line 100: seems this is always returned, so you can just return it instead of creating an optional that you set once

sfc-gh-afedorov (changes requested):
let's review what we're trying to accomplish here and how to best split up the components. specifically, let's make a PR for just the Sentry driver and have it expose just the functionality that the Sentry module exposes.

lambda_src/drivers/process_google.py
sfc-gh-afedorov: then can you undo these changes?

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: exceptions using the `sentry_sdk`.

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: Yeah that checks out. Makes more sense to be on the Snowflake side.

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: Makes sense.

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: debug level makes sense. However, for the first part we don't know for sure until it's actually invoked. Invoking code just calls the `process_row` function of the respective driver. We still want to know from the drivers PoV that the correct driver was invoked. From invoker side of the code, we can have a log saying `Invoking the {driver_name} driver.`. Is the preferable?

Does that answer the strategy question? When I was looking into the logs, I didn't know if the sentry_logger driver was correctly invoked and wanted the logs to reflect that.

lambda_src/drivers/process_google.py
sfc-gh-pkommini: The discovery stubs are available and we don't need to ignore it. for google.oauth2 stubs are not available and hence we still need to ignore.

lambda_src/drivers/process_https.py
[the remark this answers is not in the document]
sfc-gh-pkommini: The logger here was modified  to `CONSOLE_LOGGER`. Meaning we're logging in GEFF lambda logs as opposed to GEFF project in Sentry.

sfc-gh-pkommini:
@sfc-gh-afedorov Based on you feedback. I minimized the PR to only have the `sentry_logger` driver changes and the relevant `utils` and `log` dependencies. Trimmed everything else into a separate PR that will fix the logging strategy across the project.

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: moved a different PR.

lambda_src/drivers/process_sentry_logger.py
sfc-gh-pkommini: Will add invoking (from lambda_function) and invoked (from respective) for every driver in a separate PR.

lambda_src/drivers/process_https.py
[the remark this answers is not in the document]
sfc-gh-pkommini: Do you not want 400s logged at all?

lambda_src/log.py
```diff
@@ -47,3 +48,35 @@ def format_trace(e: Exception) -> str:
...
+
+    Returns:
+        Logger: Returns the logger object.
+    """
+    l = logging.getLogger(logger_name)
+    l.setLevel(level)
+    return l
+
+
+def get_loggers() -> Tuple[Logger, Logger, Logger]:
```
sfc-gh-afedorov, line 69: do we always need all three loggers? would it make sense to have the calling code just get the loggers they need instead of calling `get_loggers` which calls `getLogger` three times, of which only one might be used?

sfc-gh-pkommini: actually, u're right. We don't. let me fix.

lambda_src/drivers/process_https.py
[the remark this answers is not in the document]
sfc-gh-afedorov: correct -- when a GEFF user hits an API and gets an error response, that is the users' business and we want to handle that and not have our ops folks worrying about why a user is getting error responses. most likely they are developing and have yet to figure out how their API's auth works.

lambda_src/drivers/process_sentry_logger.py
```diff
@@ -0,0 +1,45 @@
...
+        ts (str): The timestamp of the error.
+        history_url (str): URL of the erroring object in history.
+    """
+    CONSOLE_LOGGER.info(f'sentry_logger driver invoked.')
+
+    with push_scope() as scope:
+        scope.set_extra('history_url', history_url)
+        sentry_sdk.set_tag(
+            (
+                'PIPE_NAME'
```
sfc-gh-pkommini, line 33: Starting to think PIPE_NAME, TASK_NAME and such is not needed here. @sfc-gh-afedorov I'll just receive these fields into a kwargs.

PR #56, opened 2022-06-17
Enabled Gzip Compression in the lambda response 

sfc-gh-ksharma:
Changed lambda_src.py file for enabling GZIP compressed response and show error if response is greater than payload limit of 6mb even after compression

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -8,7 +8,7 @@ from urllib.error import HTTPError, URLError
 from urllib.parse import parse_qsl, urlparse
 from urllib.request import Request, urlopen
 
-from ..utils import LOG, parse_header_links, pick
+from ..utils import LOG, parse_header_links, pick, zip
 from ..vault import decrypt_if_encrypted
 
 
@@ -126,11 +126,16 @@ def process_row(
             response_headers = dict(res.getheaders())
             res_body = res.read()
             LOG.debug(f'Got the response body with length: {len(res_body)}')
-
-            raw_response = (
-                decompress(res_body)
-                if res.headers.get('Content-Encoding') == 'gzip'
+            #Adding zip line below
+            zipped_response = (
+                zip(dumps(res_body)) 
+                if res.headers.get('Accept-Encoding') == 'gzip' 
                 else res_body
+            )
+            raw_response = (
+                decompress(zipped_response)
+                if res.headers.get('Content-Encoding') == 'gzip'
+                else zipped_response
             )
             response_date = (
                 parsedate_to_datetime(response_headers['Date']).isoformat()
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -126,16 +126,18 @@ def process_row(
             response_headers = dict(res.getheaders())
             res_body = res.read()
             LOG.debug(f'Got the response body with length: {len(res_body)}')
-            #Adding zip line below
+            """
+            #---------------Adding zip line below------------
             zipped_response = (
                 zip(dumps(res_body)) 
                 if res.headers.get('Accept-Encoding') == 'gzip' 
                 else res_body
             )
+            """
             raw_response = (
-                decompress(zipped_response)
+                decompress(res_body)
                 if res.headers.get('Content-Encoding') == 'gzip'
-                else zipped_response
+                else res_body
             )
             response_date = (
                 parsedate_to_datetime(response_headers['Date']).isoformat()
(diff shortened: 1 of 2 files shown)


lambda_src/drivers/process_https.py
```diff
@@ -126,7 +126,6 @@ def process_row(
             response_headers = dict(res.getheaders())
             res_body = res.read()
             LOG.debug(f'Got the response body with length: {len(res_body)}')
-
```
sfc-gh-afedorov, line 129: stray formatting change -- let's not change unrelated files

lambda_src/lambda_function.py
```diff
@@ -150,9 +152,14 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
-        response = {'statusCode': 200, 'body': data_dumps}
+        response = {
+            'statusCode': 200,
+            'body': base64.b64encode(gzip.compress(data_dumps.encode())).decode(),
+            'isBase64Encoded': True,
+            'headers': {'Content-Encoding': 'gzip'}
+        }
 
-    if len(response) > 6_000_000:
+    if len(response['body']) > 6_000_000:
```
sfc-gh-afedorov, line 162: this is a good catch but let's change one thing at a time in a PR and make a separate one for this change

sfc-gh-afedorov (changes requested):
this is a good start. I made a few minor comments, please address those.

also, please remove the vestigial existing gzip functionality that this is making obsolete.

finally, please include screenshots of a larger file being downloaded, e.g. [this 10MB file](https://gist.githubusercontent.com/sfc-gh-afedorov/06a17aa6a3898fd9b2e2d913efe32249/raw/74e721db61bb8e5a3f1a013ef47fd9baa2caeddf/lorem-ipsum-10mb.txt) should error if you load it in the old one and work fine if you lead it in the new one

lambda_src/lambda_function.py
```diff
@@ -5,6 +5,8 @@
 from json import dumps, loads
 from typing import Any, Dict, Text
 from urllib.parse import urlparse
+import gzip
```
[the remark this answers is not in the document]
sfc-gh-ksharma: Changed but this is not working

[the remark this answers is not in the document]
sfc-gh-ksharma: Done

sfc-gh-ksharma:
**Old Version Compression:**
Screenshot 2022-06-20 at 12 30 32 PM

**New Version:**
Screenshot 2022-06-20 at 12 55 09 PM

sfc-gh-afedorov (changes requested):
please remove the vestigial code before considering this complete

lambda_src/lambda_function.py
```diff
@@ -139,7 +141,7 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
         res_data.append(
             [
                 row_number,
-                zip(dumps(row_result)) if response_encoding == 'gzip' else row_result,
```
sfc-gh-afedorov, line 142: once this is removed, `response_encoding` becomes an unused variable. which editor do you use? ideally, it should pick up vestigial code like this.

sfc-gh-ksharma: I am using VS CODE Editor. Changed the code, deleted the unused variable

sfc-gh-afedorov: does VS Code show you anything as off? e.g. also with the "zip" import, that should now be unused?

sfc-gh-ksharma: It didn't shown any underline so not removed that, now modified it and removed zip import

sfc-gh-afedorov: any time you work with a new language you should spend some time figuring out the best tooling around it. in this case, it seems VSCode doesn't have it set up by default but you should follow [these steps](https://code.visualstudio.com/docs/python/linting) to install `pylint` so that it helps you with things like this: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unused-import.html

sfc-gh-ksharma: Sure Andrey I will remember this in Future

sfc-gh-afedorov (approved):
not sure about description but lgtm if it looks good to you @sfc-gh-pkommini

sfc-gh-pkommini:
@sfc-gh-ksharma @sfc-gh-afedorov Is this encoding meant to encoding all HTTP responses from `Lambda -> API Gateway -> EF` or just the ones where the compression on the EF is enabled?

PR #58, opened 2022-08-18
Fix destination-uri param, add strftime flags to the path

sfc-gh-nlele:
- Fix the prefix path issue and add support for strftime flags in the path. 

- Remove the data folder from the path, that was being added in absence of a prefix.

-------
Bruk had requested if we could support adding timestamps in the path, such that the path would be like `dir/dir/dir/YYYY/MM/DD/` and the files stored would have timestamps too, such as: `row-0-YYYY-MM-DD.json`

Resulting in:
`dir/dir/dir/YYYY/MM/DD/batch-id/row-0-YYYY-MM-DD.json`

-------

Test:

For destination URI: `s3://nachiket-dev-bucket/a/b/c/%Y/%d/%m/alpha/%m/beta/%y_%m_%d/gamma`

Screenshot 2022-09-15 at 4 59 28 PM

For destination URI: `s3://nachiket-dev-bucket/%m/%d/%Y`

Screenshot 2022-09-15 at 5 00 44 PM

For destination URI: `s3://nachiket-dev-bucket/` (No prefix)

Screenshot 2022-09-15 at 5 01 40 PM

For destination URI: `s3://nachiket-dev-bucket/foobar/` 

Screenshot 2022-09-15 at 5 02 20 PM

For destination URI: `s3://nachiket-dev-bucket/foobar` 

Screenshot 2022-09-15 at 5 34 21 PM

### DIFF
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -32,13 +32,9 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-        
     parts = parsed_url.path.split('/')
-    prefix_parts = [part for part in parts if not part.startswith('%') and part]
-    strftime_parts = [time.strftime(part) for part in parts if part.startswith('%')]
-
-    prefix_with_time = "/".join(prefix_parts + strftime_parts)
-    prefix = prefix_with_time if parsed_url.path.count('/') >= 1 else ''
+    prefix = "/".join([x for x in parts if x]) 
+    prefix = time.strftime(prefix) if parsed_url.path.count('/') >= 1 else ''
     LOG.debug(f'Parsed bucket = {bucket}, prefix = {prefix}.')
     return bucket, prefix
 
@@ -93,7 +89,7 @@ def initialize(destination, batch_id: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
     if not prefix:
-        prefixed_folder_path = f'{DATA_FOLDER_NAME}/{batch_id}/'
+        prefixed_folder_path = f'{batch_id}/'
     else:
         prefixed_folder_path = f'{prefix}/{batch_id}/'
 
@@ -114,7 +110,7 @@ def write(
     )
 
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{batch_id}/row-{row_index}.data.json'
     else:
         prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}.data.json'
     s3_uri = f's3://{bucket}/{prefixed_filename}'
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -16,7 +16,6 @@ AWS_REGION = os.environ[
 ]  # Placeholder while in dev TODO: change as variable/header
 S3_CLIENT = boto3.client('s3', region_name=AWS_REGION)
 MANIFEST_FILENAME = 'MANIFEST.json'
-DATA_FOLDER_NAME = 'data'
 MANIESTS_FOLDER_NAME = 'meta'
 
 
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -32,7 +32,13 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    prefix = parsed_url.path.strip('/') if parsed_url.path.count('/') >= 1 else ''
+        
+    parts = parsed_url.path.split('/')
+    prefix_parts = [part for part in parts if not part.startswith('%') and part]
+    strftime_parts = [time.strftime(part) for part in parts if part.startswith('%')]
+
+    prefix_with_time = "/".join(prefix_parts + strftime_parts)
+    prefix = prefix_with_time if parsed_url.path.count('/') >= 1 else ''
     LOG.debug(f'Parsed bucket = {bucket}, prefix = {prefix}.')
     return bucket, prefix
 
@@ -106,12 +112,11 @@ def write(
         if isinstance(datum, list)
         else json.dumps(datum, default=str)
     )
-    date = time.strftime("%Y-%m-%d")
-
+    
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}-{date}.data.json'
+        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
     else:
-        prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}-{date}.data.json'
+        prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}.data.json'
     s3_uri = f's3://{bucket}/{prefixed_filename}'
 
     return {
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -30,10 +30,10 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     """
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
-    bucket = parsed_url.netloc
-    prefix = strftime(parsed_url.path).lstrip('/') if parsed_url.path.count('/') >= 1 else ''
-    LOG.debug(f'Parsed bucket = {bucket}, prefix = {prefix}.')
-    return bucket, prefix
+    return (
+        parsed_url.netloc,
+        strftime(parsed_url.path[1:])  # remove leading slash
+    )
 
 
 def estimated_record_size(records: List[Dict[Text, Any]]) -> float:
@@ -82,17 +82,19 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
     )
 
 
-def initialize(destination: Text):
+def initialize(destination: Text, batch_id: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
     prefixed_folder_path = ''
     if prefix and prefix.count('/') and not prefix.endswith('/'):
-        prefixed_folder_path = '/'.join(prefix.split('/')[:-1]) + '/'
+        # Removes characters after the last '/' in a path, since we want to initilaize a folder
+        # E.g. /a/b/c -> /a/b/
+        prefixed_folder_path = '/'.join(prefix.split('/')[:-1]) + '/' 
     elif prefix.endswith('/'):
         prefixed_folder_path = prefix
 
     if prefixed_folder_path:
-        return write_to_s3(bucket, prefixed_folder_path, content)
+        write_to_s3(bucket, prefixed_folder_path, content)
 
 
 def write(
@@ -108,12 +110,7 @@ def write(
         else json.dumps(datum, default=str)
     )
 
-    if not prefix:
-        prefixed_filename = f'{batch_id}-row-{row_index}.data.json'
-    elif prefix.endswith('/'):
-        prefixed_filename = f'{prefix}{batch_id}-row-{row_index}.data.json'
-    else:
-        prefixed_filename = f'{prefix}-{batch_id}-row-{row_index}.data.json'
+    prefixed_filename = f'{prefix}{batch_id}_row_{row_index}.data.json'
     s3_uri = f's3://{bucket}/{prefixed_filename}'
 
     return {
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -3,6 +3,7 @@ import os
 from random import sample
 from typing import Any, AnyStr, Dict, Generator, List, Optional, Text, Tuple, Union
 from urllib.parse import urlparse
+import time
 
 import boto3
 from botocore.exceptions import ClientError
@@ -31,7 +32,7 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    prefix = parsed_url.path.split('/', 2)[1] if parsed_url.path.count('/') >= 2 else ''
+    prefix = parsed_url.path.strip('/') if parsed_url.path.count('/') >= 1 else ''
     LOG.debug(f'Parsed bucket = {bucket}, prefix = {prefix}.')
     return bucket, prefix
 
@@ -105,11 +106,12 @@ def write(
         if isinstance(datum, list)
         else json.dumps(datum, default=str)
     )
+    date = time.strftime("%Y-%m-%d")
 
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}-{date}.data.json'
     else:
-        prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}-{date}.data.json'
     s3_uri = f's3://{bucket}/{prefixed_filename}'
 
     return {
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -112,7 +112,7 @@ def write(
         if isinstance(datum, list)
         else json.dumps(datum, default=str)
     )
-    
+
     if not prefix:
         prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
     else:
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -88,7 +88,7 @@ def initialize(destination: Text, batch_id: Text):
     content = ''  # We use empty body for creating a folder
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
-    prefixed_folder_path = re.sub(r'/[^/]*$', "/", prefix)
+    prefixed_folder_path = re.sub(r'/[^/]*$', "/", prefix) if '/' in prefix else ''
 
     if prefixed_folder_path:
         write_to_s3(bucket, prefixed_folder_path, content)
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -88,10 +88,10 @@ def initialize(destination: Text, batch_id: Text):
     content = ''  # We use empty body for creating a folder
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
-    prefixed_folder_path = re.sub(r'/[^/]*$', "/", prefix) if '/' in prefix else ''
+    prefix_folder = re.sub(r'/[^/]*$', "/", prefix) if '/' in prefix else ''
 
-    if prefixed_folder_path:
-        write_to_s3(bucket, prefixed_folder_path, content)
+    if prefix_folder:
+        write_to_s3(bucket, prefix_folder, content)
 
 
 def write(
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -31,9 +31,7 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    parts = strftime(parsed_url.path).split('/')
-    prefix = "/".join([x for x in parts if x])
-    prefix = prefix if parsed_url.path.count('/') >= 1 else ''
+    prefix = strftime(parsed_url.path).lstrip('/') if parsed_url.path.count('/') >= 1 else ''
     LOG.debug(f'Parsed bucket = {bucket}, prefix = {prefix}.')
     return bucket, prefix
 
@@ -84,15 +82,17 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
     )
 
 
-def initialize(destination, batch_id: Text):
+def initialize(destination: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
-    if not prefix:
-        prefixed_folder_path = f'{batch_id}/'
-    else:
-        prefixed_folder_path = f'{prefix}/{batch_id}/'
+    prefixed_folder_path = ''
+    if prefix and prefix.count('/') and not prefix.endswith('/'):
+        prefixed_folder_path = '/'.join(prefix.split('/')[:-1]) + '/'
+    elif prefix.endswith('/'):
+        prefixed_folder_path = prefix
 
-    return write_to_s3(bucket, prefixed_folder_path, content)
+    if prefixed_folder_path:
+        return write_to_s3(bucket, prefixed_folder_path, content)
 
 
 def write(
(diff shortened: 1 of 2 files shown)


sfc-gh-afedorov (changes requested):
I think there's a simpler solution to this problem -- if we just add strftime format codes (from the 1989 C standard), most folks will know what that means and we can just mention that and link [these Python docs](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior) in our explanation of it

lambda_src/drivers/destination_s3.py
```diff
@@ -31,7 +32,13 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    prefix = parsed_url.path.split('/', 2)[1] if parsed_url.path.count('/') >= 2 else ''
+        
+    parts = parsed_url.path.split('/')
+    prefix_parts = [part for part in parts if not part.startswith('%') and part]
+    strftime_parts = [time.strftime(part) for part in parts if part.startswith('%')]
+
+    prefix_with_time = "/".join(prefix_parts + strftime_parts)
```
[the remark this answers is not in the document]
sfc-gh-pkommini: @sfc-gh-nlele There is a piece from our discussion misinterpreted here:

There are 3 parts to the prefix from the user 's3://bucket-name/<optional resource path>/<optional-date-format-string>' here:
1. Bucket name
2. The optional prefix path may or may not exist (if missing the data should go to the `data/ <optional-date-format-string>` path rather than directly in the bucket, if present then data goes into for e.g.: `bucket-name/a/b/c/%y/%m/%d/` )
3. The optional time format string also may or may not exist. If present data goes into e.g. `bucket-name/a/b/c/%y/%m/%d/`, if not then it goes into the prefix path, e.g. `s3://bucket-name/a/b/c`
4. If both the optional resource prefix path and date format string are missing then we just place the data in the `s3://bucket-name/data/` path.

@sfc-gh-afedorov Does this make sense?

sfc-gh-pkommini (changes requested):

lambda_src/drivers/destination_s3.py
```diff
@@ -31,7 +32,13 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
...
+    prefix = prefix_with_time if parsed_url.path.count('/') >= 1 else ''
```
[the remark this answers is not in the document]
sfc-gh-nlele: `s3://dev-bucket/geff-function-3-results` 

Considering this to be the destination URI, it wouldn't be actually, since `urlparse` returns `/geff-function-3-results` as the path, which has a single `/`.

lambda_src/drivers/destination_s3.py
```diff
@@ -31,7 +32,13 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    prefix = parsed_url.path.split('/', 2)[1] if parsed_url.path.count('/') >= 2 else ''
```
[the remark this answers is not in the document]
sfc-gh-nlele: For something like `s3://dev-bucket/foo`, this would return an empty string, since it'd expect at least two `/`s, but the path (`/foo`) would only have one `/`. 

And for a nested path such as this, `s3://dev-bucket/a/b/c/d`, the prefix path returned, would be `a`, since the split would be: `['', 'a', 'b/c/d']`.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: Yes, I see that from the code, but all these rules seem to add avoidable extrinsic complexity and most importantly won't support the thing that the user likely wants, which is what AWS does, data stored in e.g. —

~~~
/AWSLogs/123456789012/CloudTrail/ap-northeast-1/2020/03/02/123456789012_CloudTrail_ap-northeast-1_20200302T0020Z_Fo891lLUhaxnyLhj1.json.gz
~~~

in order to support this, we could allow the user to specify a prefix using `strftime` flags and parameters specifying account id (`{1}`) and region (`{2}`) like —

~~~
/AWSLogs/{1}/CloudTrail/{2}/%Y/%m/%d/{1}_CloudTrail_{2}_%Y%m%dT%H%MZ_
~~~

Not sure if we already support the parameters, but I'm OK with adding `strftime` first and then the parameters in a separate PR, if preferred.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: what is the intent of this code, then? is it to handle the case where the user provides a destination that's just a simple string like `"dev-bucket"`?

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: Ah, I see, so this code only intended for there to be a single folder in `parsed_url.path` instead of a nested URL?

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-pkommini: All Bruk wanted to the year, month, date options in the path and that was it. Additionally he said it would be good to have it in the file name but he was okay with not having it also.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-pkommini: default prefix for data should be `/data` and not empty. So that we don't end up with a lot of files in the bucket home dir in the UI.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-nlele: Yess, only the first level in the path would be created with the code.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: k, we can skip the parametrization for now if you'd like, but let's let the user put the strftime flags wherever they want...

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: in the case of a destination prefix, having a default seems to be non-obvious. if someone says `s3://foo/` do we add the default or did they explicitly want the root by ending it with a `/`? is that something we should document, that there's a difference between this and `s3://foo`, which really means `s3://foo/data`?

I think the easiest thing if we have a `destination-uri` is just to respect what the user requests -- put the data there in whatever format makes sense, with an optional translation layer via parameters and strftime flags, etc.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-pkommini: Where they want meaning before or after an optional prefix? So what you're saying is it's okay to have something like this?

```
'prefix' = 's3://my-bucket/a/%y/b/%d/%m/c'

-- or

'prefix' = 's3://my-bucket/a/%Y/b/%d/c/%m'
```

This would make the logic to collate the sfrftime variables and passing it to the strftime function unnecessarily complicated.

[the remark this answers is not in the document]
sfc-gh-pkommini: it wudn't be to complicated if we apply it to the whole string I guess. @sfc-gh-nlele Could you please test strftime'ing the full path and see if that allows us to keep our logic simple?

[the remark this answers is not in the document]
sfc-gh-pkommini: actually just tested and it seems to work just fine.
image

lambda_src/drivers/destination_s3.py
```diff
@@ -31,7 +32,13 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
...
+        
+    parts = parsed_url.path.split('/')
```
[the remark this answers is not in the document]
sfc-gh-afedorov: checking in 7 days later -- what's up?

lambda_src/drivers/destination_s3.py
```diff
@@ -107,7 +110,7 @@ def write(
     )
 
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
```
sfc-gh-afedorov, line 110: if we delete the references to `DATA_FOLDER_NAME`, we should delete the unused variable up top, as well. please set up pylint to tell you about things like this https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unused-variable.html

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-nlele: I thought we could strftime the string on line 36, after it has the expected format, like from `/a/b/c/` to `a/b/c`. But will strftime it before the split if it is a better pattern.

lambda_src/drivers/destination_s3.py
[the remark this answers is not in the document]
sfc-gh-afedorov: yeah, I think `strftime` is kind of like `sprintf` [1] in that it is intended to print the whole string to another string but with certain flags replaced. I'm not 100% sure if there are times when someone might not want this replacement, but it seems easy enough to make it optional via something like a `destination-uri-raw` option, if it becomes necessary for some reason...

1. https://www.geeksforgeeks.org/sprintf-in-c/

lambda_src/drivers/destination_s3.py
```diff
@@ -31,9 +31,9 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    parts = parsed_url.path.split('/')
-    prefix = "/".join([x for x in parts if x]) 
-    prefix = time.strftime(prefix) if parsed_url.path.count('/') >= 1 else ''
+    parts = strftime(parsed_url.path).split('/')
+    prefix = "/".join([x for x in parts if x])
```
sfc-gh-afedorov, line 35: what is the purpose of this `split` followed by a `filter` and a `join`? how would you explain this to users in the documentation? perhaps I'm missing something, but I would expect the logic here to be more straightforward, e.g. generate the file name to which to write with `{strftime(destination_uri)}-{batch_id}-{row}.json.gz`, unless `destination_uri` ends with a `/` in which case skip the first `-` and just put it in the folder specified by the user.

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +82,17 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
     )
 
 
-def initialize(destination, batch_id: Text):
```
sfc-gh-nlele, line 85: Removed the `batch_id` arg, as it becomes vestigial since we no longer use it while initializing the folder

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +82,17 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
...
+def initialize(destination: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
-    if not prefix:
-        prefixed_folder_path = f'{DATA_FOLDER_NAME}/{batch_id}/'
-    else:
-        prefixed_folder_path = f'{prefix}/{batch_id}/'
+    prefixed_folder_path = ''
+    if prefix and prefix.count('/') and not prefix.endswith('/'):
```
sfc-gh-nlele, line 89: 1) `not prefix.endswith('/')` for when the `destination-uri` is: `s3://bucket_name/a/b/c`
2) `prefix.count('/')`  for when the `destination-uri` is: `s3://bucket_name/a`, since we don't want a folder in this case. The `/` from `/a` gets lstripped in line 34, so the count becomes zero.

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +82,17 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
...
+        prefixed_folder_path = '/'.join(prefix.split('/')[:-1]) + '/'
+    elif prefix.endswith('/'):
+        prefixed_folder_path = prefix
 
-    return write_to_s3(bucket, prefixed_folder_path, content)
+    if prefixed_folder_path:
```
sfc-gh-nlele, line 94: We had a `prefixed_folder_path` earlier for an empty prefix, and now, when the prefix is empty, `write_to_s3(bucket, prefixed_folder_path, content)` would error. 

Added a condition to check `prefixed_folder_path`.

lambda_src/drivers/destination_s3.py
```diff
@@ -31,7 +31,7 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
     bucket = parsed_url.netloc
-    prefix = parsed_url.path.split('/', 2)[1] if parsed_url.path.count('/') >= 2 else ''
+    prefix = strftime(parsed_url.path).lstrip('/') if parsed_url.path.count('/') >= 1 else ''
```
sfc-gh-nlele, line 34: `lstrip` since we don't the prefixes to be of the form `/a/b/c` , and instead, `a/b/c` when they reach the write functions.

sfc-gh-pkommini: *don't want?

sfc-gh-nlele: Yess.

sfc-gh-pkommini:
@sfc-gh-nlele The testing looks good. Could you also test both sync and async flows as well?

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: this is a generic API so just because it's vestigial here doesn't mean it won't be useful in another destination driver

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: why does this need to happen? why does anything need to happen when initializing an S3 destination?

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: the return here seems to suggest there's meaning to this response value, but that's misleading because the code that invokes it does not store the return value

lambda_src/lambda_function.py
```diff
@@ -44,7 +44,7 @@ def async_flow_init(event: Any, context: Any) -> Dict[Text, Any]:
         f'geff.drivers.destination_{urlparse(destination).scheme}'
     )
     # Ignoring style due to dynamic import
-    destination_driver.initialize(destination, batch_id)  # type: ignore
```
sfc-gh-afedorov, line 47: this is a generic API so just because it's vestigial here doesn't mean it won't be useful in another destination driver

lambda_src/drivers/destination_s3.py
```diff
@@ -107,9 +109,11 @@ def write(
     )
 
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{batch_id}-row-{row_index}.data.json'
```
sfc-gh-afedorov, line 112: this logic repeats regardless of which of the three options is true. instead, you can set this variable and then re-use it in the below.

lambda_src/drivers/destination_s3.py
```diff
@@ -107,9 +109,11 @@ def write(
...
+    elif prefix.endswith('/'):
+        prefixed_filename = f'{prefix}{batch_id}-row-{row_index}.data.json'
     else:
-        prefixed_filename = f'{prefix}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{prefix}-{batch_id}-row-{row_index}.data.json'
```
sfc-gh-afedorov, line 116: no need to separate `prefix` and `batch_id` on the users' behalf (they can do it if they want), i.e. we can collapse these two cases

lambda_src/drivers/destination_s3.py
```diff
@@ -107,9 +109,11 @@ def write(
     )
 
     if not prefix:
-        prefixed_filename = f'{DATA_FOLDER_NAME}/{batch_id}/row-{row_index}.data.json'
+        prefixed_filename = f'{batch_id}-row-{row_index}.data.json'
+    elif prefix.endswith('/'):
+        prefixed_filename = f'{prefix}{batch_id}-row-{row_index}.data.json'
```
sfc-gh-afedorov, line 114: probably best to separate batch and row with `_` instead of `-row-`, that's how AWS does it

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +82,17 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
...
-        prefixed_folder_path = f'{prefix}/{batch_id}/'
+    prefixed_folder_path = ''
+    if prefix and prefix.count('/') and not prefix.endswith('/'):
+        prefixed_folder_path = '/'.join(prefix.split('/')[:-1]) + '/'
+    elif prefix.endswith('/'):
+        prefixed_folder_path = prefix
 
-    return write_to_s3(bucket, prefixed_folder_path, content)
+    if prefixed_folder_path:
+        return write_to_s3(bucket, prefixed_folder_path, content)
```
sfc-gh-afedorov, line 95: the return here seems to suggest there's meaning to this response value, but that's misleading because the code that invokes it does not store the return value

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: `parsed_url.path.count('/') >= 1` seems identical to `'/' in parsed_url.path`, but I'm not able to find a string that would return a non-empty `path` that doesn't start with `/`, so this seems like like extra complexity that doesn't do anything in practice

looking at fundamentals here and as you said @sfc-gh-nlele, the difference between what `urlparse`'s `path` returns and what `s3`'s `put_object` takes is a single character, the always-present `/` prefix in a non-empty `path` that's implicit in `put_object`, so we don't want to strip anything more.

if you think about it more, that's sensible adjustment from general case: removing the slash makes `s3://foo` and `s3://foo/` parse to the same thing, and that's fine for our use case and also not the general case that `urlparse` authors wanted, hence the adapter makes sense.

seems this gets simple enough that we can skip printing logs and then the function boils down to —

~~~python
u = urlparse(destination)
return (
  u.netloc,
  strftime(u.path[1:])  # remove leading slash
)
~~~

sfc-gh-afedorov (changes requested):
getting close, altho I'm a little confused why we need to do anything in `destination_driver.initialize` and definitely don't think we should be changing the general API as part of this request. not against changing it, it just doesn't seem related to this specific fix.

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Ah okay, will change it.

lambda_src/lambda_function.py
sfc-gh-nlele: I'd considered only the current s3 destination driver, will add it back for the future drivers.

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Sure, will remove remove that case.

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Ah right, should this then just be made into an invocation by removing the return?

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Makes sense. Will make the change.

Could the logs for the bucket and the prefix be an extra piece of information for the user that they could use while debugging and such?

sfc-gh-afedorov: That's something we should decide as part of the logging strategy (GSE-2105), but probably not -- the user knows what destination they passed in, and since the transformation is now relatively trivial, there's no need to log it.

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: Not 100% certain but from my read of it —

~~~
def initialize(destination, batch_id):
    pass
~~~

sfc-gh-pkommini (changes requested):

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: > That's something we should decide as part of the logging strategy (GSE-2105), but probably not -- the user knows what destination they passed in, and since the transformation is now relatively trivial, there's no need to log it.

Ah okay 👍

sfc-gh-nlele: > I don't think this is the behavior we want here?
> 
> image
> 
> Do you want prefix to be `'a/b'` or `''` in this case?

The value that gets passed through `parsed_url.path` will always be in the form of `/a/b` if it exists right? Like for the URI `s3://dev-bucket/a/b`  


Which would result in `a/b` when lstripped

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Sure, will add comments for it.

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: yes, I think these three can be collapsed into a single case

lambda_src/lambda_function.py
sfc-gh-afedorov: this seems outside the scope of this PR and could be troublesome to collaborators. not sure if that's the case here, but there might be other drivers in forks or branches of this code and removing this line would prevent them from applying this change to their versions of the code.

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: yes, but there might be more than one slash if for whatever reason a prefix path starts with a `/`

if you'd like to add an `assert prefix[0] == '/'` that would ensure that the code works as exepcted

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +83,15 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
...
+def initialize(destination: Text, batch_id: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
-    if not prefix:
-        prefixed_folder_path = f'{DATA_FOLDER_NAME}/{batch_id}/'
-    else:
-        prefixed_folder_path = f'{prefix}/{batch_id}/'
+    # Regex captures characters after and including the rightmost '/' in a path,
+    # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
+    prefixed_folder_path = re.sub(r'/[^/]*$', "/", prefix) if '/' in prefix else ''
```
sfc-gh-nlele, line 91: Added this conditional: ` if '/' in prefix else ''` for the case: `s3://dev-bucket/a`

The value of `prefix` will be `a`, and without the conditional, an empty object `a` will be created in the bucket, as it wouldn't get replaced/removed.

sfc-gh-afedorov: seems like regex handles the empty case ok --

~~~python
In [1]: import re

In [2]: prefix = ''

In [3]: re.sub(r'/[^/]*$', "/", prefix)
Out[3]: ''
~~~

sfc-gh-afedorov: could we call this `prefix_folder` since this is part of the prefix, not a prefixed something else.

sfc-gh-nlele: True, the empty case works okay, but a path like `s3://dev-bucket/foobar` would be problematic, since the value of `prefix` will be `foobar`, which when passed through the regex operation will result in `foobar` as the folder path. 

 
Screenshot 2022-09-15 at 8 04 52 PM

sfc-gh-nlele: > could we call this `prefix_folder` since this is part of the prefix, not a prefixed something else?

Sure 👍

lambda_src/drivers/destination_s3.py
```diff
@@ -82,15 +83,15 @@ def write_to_s3(bucket: Text, filename: Text, content: AnyStr) -> Dict[Text, Any
...
+def initialize(destination: Text, batch_id: Text):
     bucket, prefix = parse_destination_uri(destination)
     content = ''  # We use empty body for creating a folder
-    if not prefix:
-        prefixed_folder_path = f'{DATA_FOLDER_NAME}/{batch_id}/'
-    else:
-        prefixed_folder_path = f'{prefix}/{batch_id}/'
+    # Regex captures characters after and including the rightmost '/' in a path,
+    # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
+    prefix_folder = re.sub(r'/[^/]*$', "/", prefix) if '/' in prefix else ''
```
sfc-gh-afedorov, line 91: replace with `''` and then add the trailing slash when calling `write_to_s3`?

sfc-gh-afedorov, line 91: could we remove the `if` piece since regex replace on an empty string will work regardless?

sfc-gh-afedorov (approved):

sfc-gh-afedorov (approved):
lgtm

sfc-gh-pkommini (approved):
lgtm!

PR #59, opened 2022-09-27
Add request locking mechanism

sfc-gh-nlele:
When a request takes more than 30 seconds in a synchronous flow, the API gateway times out. Snowflake can then make a repeated request, since the result of the original request is not known. 

This PR adds a way to not repeat requests if the gateway times out, refactors some of the batch processing functions and returns additional error responses as a way for the user to get more information about their request's status.

Tests:

- **API response time: 35s, Sync**
    No repeated requests were made.
Screenshot 2023-01-04 at 11 01 37 PM

    Output in Snowflake:
Screenshot 2023-01-04 at 11 01 05 PM

More cases:

- **API response time: 15s, Async**
    No entries added in DynamoDB, since the request is async.
    Output in Snowflake:

Screenshot 2023-01-04 at 10 34 10 PM

- **API response time: 10s, Sync, Response Size: 3.4 MB**
    Response not written in DynamoDB, since the response time is <20s.
    Output in Snowflake:

Screenshot 2023-01-04 at 10 49 13 PM

- **API response time: 35s, Sync, Response size:  759 KB**

    Output in Snowflake:

Screenshot 2023-01-04 at 8 53 13 PM

- **API response time: 70s, Sync**

   Output in Snowflake:

Screenshot 2023-01-04 at 11 06 23 PM


lambda_src/lambda_function.py
```diff
@@ -99,13 +102,23 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
     if write_uri:
         destination_driver = import_module(
             f'geff.drivers.destination_{urlparse(write_uri).scheme}'
         )
+    else:
+        if not batch_id_exists:
+            initialize_dynamodb_item(batch_id) # For the first Lambda
+        else:
+            while get_response(batch_id) is None: # For the subsequent Lambda
+                sleep(5)
```
sfc-gh-pkommini, line 118: what do we need `sleep` here?

sfc-gh-nlele: Had added it to keep checking for a response in the table, in case a response has not been returned by the time the second Lambda is triggered.

sfc-gh-nlele: Had added it to keep checking for a response in the table, in case a response has not been returned by the time the second Lambda is triggered.

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
...
+    response = item['Item']['response']
+    return response
+
+
+def check_if_initialized(batch_id: Text):
+    """
+    Check if a response for a batch id exists in the table
+    """
+    item = table.get_item(Key={'batch_id': batch_id})
+    try:
```
sfc-gh-afedorov, line 42: unless there's a specific reason (e.g. a busy code path with the exception very rare) use an `if` here instead of `try`/`catch`

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
...
+DYNAMODB_TABLE = 'geff-requests'
+DYNAMODB_RESOURCE = boto3.resource('dynamodb', region_name=AWS_REGION)
+table = DYNAMODB_RESOURCE.Table(DYNAMODB_TABLE)
+
+
+def write_dynamodb_item(batch_id: Text, response: Dict):
+    """
+    Write to the requests table, a batch id and a response
+    """
+    table.put_item(Item={'batch_id': batch_id, 'response': response})
```
sfc-gh-afedorov, line 18: add an expiration ttl key

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
+import os
+from typing import Dict, Text
+
+import boto3
+
+AWS_REGION = os.environ[
+    'AWS_REGION'
+]
+DYNAMODB_TABLE = 'geff-requests'
```
sfc-gh-afedorov, line 9: users might want to customize the table they use for locks. by default let's not call it "GEFF requests", as it's up to the driver code whether there's a single request, or multiple requests, or something else entirely here (e.g. sending an email is usually called a request).

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
+import os
+from typing import Dict, Text
+
+import boto3
+
+AWS_REGION = os.environ[
```
sfc-gh-afedorov, line 6: put on one line (use black), and we might want to set a default when the envar is not present

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def initialize_dynamodb_item(batch_id: Text):
+    """
+    Initialize an item in the requests table with a null response
+    """
+    table.put_item(Item={'batch_id': batch_id, 'response': None})
+
+
+def get_response(batch_id: Text):
```
sfc-gh-afedorov, line 28: I think this will always be the same as the `check` function below, would it make sense to combine them?

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
...
+    return response
+
+
+def check_if_initialized(batch_id: Text):
+    """
+    Check if a response for a batch id exists in the table
+    """
+    item = table.get_item(Key={'batch_id': batch_id})
+    try:
+        response = item['Item']['response']
```
sfc-gh-afedorov, line 43: when you initialize, the response is set to `None`. are you sure it will not return `None` here? add a comment explaining this, if so.

lambda_src/response_storage.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Dict, Text
+
+import boto3
+
+AWS_REGION = os.environ[
+    'AWS_REGION'
+]
+DYNAMODB_TABLE = 'geff-requests'
+DYNAMODB_RESOURCE = boto3.resource('dynamodb', region_name=AWS_REGION)
+table = DYNAMODB_RESOURCE.Table(DYNAMODB_TABLE)
```
sfc-gh-afedorov, line 11: how will this table be created?

sfc-gh-afedorov (changes requested):
I think all the comments are here, but let's maybe pair after you address these, just to make sure the process is a little more clear

lambda_src/response_storage.py
sfc-gh-nlele: Will do 👍

lambda_src/response_storage.py
sfc-gh-nlele: Yess, have only kept one get function now.

lambda_src/lambda_function.py
```diff
@@ -7,9 +7,12 @@
 from json import dumps, loads
 from typing import Any, Dict, Text
 from urllib.parse import urlparse
+from time import sleep
 
 from .log import format_trace
 from .utils import LOG, create_response, format, invoke_process_lambda
+from .response_storage import (initialize_dynamodb_item, write_dynamodb_item,
```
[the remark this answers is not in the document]
sfc-gh-nlele: Makes sense, renamed the module, have made the functions generic too.

lambda_src/lambda_function.py
```diff
@@ -99,13 +102,23 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
     write_uri = headers.get('write-uri')
     LOG.debug(f'sync_flow() received destination: {write_uri}.')
 
     if write_uri:
         destination_driver = import_module(
             f'geff.drivers.destination_{urlparse(write_uri).scheme}'
         )
+    else:
+        if not batch_id_exists:
+            initialize_dynamodb_item(batch_id) # For the first Lambda
```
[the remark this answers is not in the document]
sfc-gh-nlele: Have added a ` if request_locking:` for this, where `request_locking` provides the backend to use, like `dynamodb` and if the header is not present, we just have a normal flow.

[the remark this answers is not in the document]
sfc-gh-nlele: > keep a common interface -- the name of the implementation should not be in the interface

Have this in a local commit actually, will push it soon.

lambda_src/lambda_function.py
```diff
@@ -99,13 +102,23 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        else:
+            while get_response(batch_id) is None: # For the subsequent Lambda
+                sleep(5)
+            if get_response(batch_id):
```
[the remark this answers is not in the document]
sfc-gh-nlele: Yess, changed this in the review with Prasanth 👍

lambda_src/lambda_function.py
```diff
@@ -157,6 +170,7 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
             'isBase64Encoded': True,
             'headers': {'Content-Encoding': 'gzip'}
         }
+        write_dynamodb_item(batch_id, response)
```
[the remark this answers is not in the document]
sfc-gh-nlele: Will do 👍

lambda_src/response_storage.py
sfc-gh-nlele: Ah okay 👍

lambda_src/response_storage.py
sfc-gh-nlele: Would it be okay to have it passed as an environment variable?

lambda_src/response_storage.py
sfc-gh-nlele: Have added resources for it [here](https://github.com/Snowflake-Labs/terraform-snowflake-api-integration-with-geff-aws/pull/71/files)

lambda_src/response_storage.py
sfc-gh-nlele: Actually, I'd considered that it was fine when it did too, because the I wanted to mainly check for the existence of the batch_id key.

sfc-gh-nlele: and if the key existed, we'd get `True` returned

lambda_src/response_storage.py
sfc-gh-nlele: oo okay 👍

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: let's not parametrize it just yet, as we have no way of justifying the ned for a user-facing interface just yet.

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: it should be enabled by default, but the user should have a way to disable it

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: cool, sorry crossing streams and missing that this wasn't ready for review yet

lambda_src/response_storage.py
sfc-gh-afedorov: not 100% sure but I think that makes sense -- and the user sets it in Terraform?

lambda_src/response_storage.py
sfc-gh-afedorov: what does `get_item` return when the key is not found? is there a better way to check?

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-nlele: Do you think it might be safer to have it disabled by default, and then enabling it with an optional header? (mainly because of the size limit)

```
....
(other headers)
request-locking = 'dynamodb'
(other headers)
...
```

lambda_src/response_storage.py
sfc-gh-nlele: It returns a `None`. 

The other way is to use chained `get`s:

```
item.get("Item", {}).get("response")
```

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        data_dumps = dumps({'data': res_data}, default=str)
+        response = {
+            'statusCode': 200,
+            'body': b64encode(compress(data_dumps.encode())).decode(),
+            'isBase64Encoded': True,
+            'headers': {'Content-Encoding': 'gzip'},
+        }
+        end_time = timer()
+        if response and (end_time - start_time) > 20:
+            LOG.debug('Storing the response in DynamoDB.')
```
sfc-gh-afedorov, line 146: store response size

sfc-gh-afedorov: @sfc-gh-bkou can advise on what exactly is required as part of his "logging/metrics strategy" task -- probably batch id and number of bytes stored?

sfc-gh-pkommini:
@sfc-gh-nlele I see that this is still in draft, is it ready to review yet?

sfc-gh-nlele:
Yes, it was. Removed the draft status 👍

sfc-gh-pkommini (dismissed):
lgtm

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+                    },
+                ]
+                for rn, *args in req_body['data']
+            ]
+        }
+    )
+    return response
+
+
+def process_request(
```
sfc-gh-afedorov, line 192: Functions that are called once should encapsulate meaning, so try to avoid vague functions like "process request". I'm a little confused which "request" you are referring to here... it seems misleading because of the things that might be called a "request" here, this is going to be potentially processing multiple rows per batch and each of those rows might have multiple RPC calls each of which might be represented by e.g. an HTTP request. Instead, let's call this `process_batch` and change the params a bit —

1. pull `batch_data` out of the body
2. filter out `headers` that start with `sf-custom-` and do the string replacements on them here
3. parse the `path` out of the `event`

Also, since this is only called once, you don't need to have default params that are always present.

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+                if ce.response['Error']['Code'] == 'ValidationException':
+                    LOG.error(ce)
+                    pass
+
+    if len(response) > 6_000_000:
+        response = response_size_error(response, req_body)
+    return response
+
+
+def response_size_error(
```
sfc-gh-afedorov, line 159: Functions should be verbs describing what they do, so let's maybe call it `construct_size_error_response`? 

not sure if it should be part of this PR or a separate task, but if you're looking to refactor this piece, there appears to be a bug here that might be worth fixing in that `response` should probably consistently be a dict constructed by `create_response` with an appropriate type signature (i.e. keys `statusCode`, `body`, `isBase64Encoded`, `headers`)

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+    response_input: Dict[Text, Any], req_body: Dict[Text, Any]
+) -> str:
+    """
+    Creates a new response object with an error message,
+    for when the response size is likely to exceed the allowed payload size.
+
+    Args:
+        response (Dict[Text, Any]): Response object to calculate the size.
```
sfc-gh-afedorov, line 167: `response_input` seems like an undescriptive name here -- it's input because it's the arg to a function?

it's a string, not a dict.

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        req_body (Any): Body of the request, obtained from the events object.
+
+    Returns:
+        Dict[Text, Any]: Represents the response with the error message.
```
sfc-gh-afedorov, line 171: I'm not sure if it's useful to repeat the type signature in the docstring format, but if you want to keep it, it returns a string, not a dict

sfc-gh-afedorov (changes requested):
big step in the right directly, let's do one more iteration and then can review the test strategy

lambda_src/drivers/process_https.py
```diff
@@ -2,7 +2,6 @@
 from email.utils import parsedate_to_datetime
 from gzip import decompress
 from json import JSONDecodeError, dumps, loads
-from re import match
```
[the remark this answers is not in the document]
sfc-gh-nlele: Will make a new PR for the unused import 👍

lambda_src/lambda_function.py
sfc-gh-nlele: > Functions should be verbs describing what they do, so let's maybe call it construct_size_error_response?

Makes sense, will change it. 

> not sure if it should be part of this PR or a separate task, but if you're looking to refactor this piece, there appears to be a bug here that might be worth fixing in that response should probably consistently be a dict constructed by create_response with an appropriate type signature (i.e. keys statusCode, body, isBase64Encoded, headers)

It seems like `create_response` currently creates a dict with only `statusCode` and `body`, and the dict that does have `isBase64Encoded` and `headers`, is manually [created](https://github.com/Snowflake-Labs/geff/blob/2aa38df9a4345e40e53a6ee4dc261dae66be3398/lambda_src/lambda_function.py#L138). Would we want to maybe add `isBase64Encoded` and `headers` as optional args in `create_response` and use it?

(Also, after making the args optional, a replacement with the function [here](https://github.com/Snowflake-Labs/geff/blob/2aa38df9a4345e40e53a6ee4dc261dae66be3398/lambda_src/lambda_function.py#L85) and [here](https://github.com/Snowflake-Labs/geff/blob/2aa38df9a4345e40e53a6ee4dc261dae66be3398/lambda_src/lambda_function.py#L88) maybe?)

lambda_src/lambda_function.py
sfc-gh-nlele: Yess, will update the docstring, meant to write `str` there actually. Since the other functions describe the type sig. as well, I thought it could be consistent with them, but can remove it if that's preferred.

lambda_src/lambda_function.py
sfc-gh-nlele: > response_input seems like an undescriptive name here -- it's input because it's the arg to a function?

Was thinking something along the lines of that, will think of a more descriptive name.

> it's a string, not a dict.

Sorry, but it seems like it's a dict? ([1](https://github.com/Snowflake-Labs/geff/blob/e601e97a1aa90649ebc18f15a66cae1a1b8e8a6b/lambda_src/lambda_function.py#L154), [2](https://github.com/Snowflake-Labs/geff/blob/2aa38df9a4345e40e53a6ee4dc261dae66be3398/lambda_src/drivers/destination_s3.py#L133))

sfc-gh-afedorov: The one that returns `statusCode` is the correct one to return to the lambda runtime. this function returns a string, and `finalize` returns a dictionary that is not properly formatted and so should be erroring when the async call completes. the best way to fix this is probably by using [TypedDict](https://docs.python.org/3.8/library/typing.html#typing.TypedDict). up to you if you want to do that refactoring here, tho -- keeping things inline as they are seems fine, as well.

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+            'isBase64Encoded': True,
+            'headers': {'Content-Encoding': 'gzip'},
+        }
+        end_time = timer()
+        if response and (end_time - start_time) > 20:
+            LOG.debug('Storing the response in DynamoDB.')
+            try:
+                close_lock(batch_id, response)  # write the response
+            except ClientError as ce:
+                if ce.response['Error']['Code'] == 'ValidationException':
```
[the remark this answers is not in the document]
sfc-gh-nlele: It's the exception returned when the size limit of an itme is hit. The error message:
> An error occurred (ValidationException) when calling the PutItem operation: Item size has exceeded the maximum allowed size

The rest of the flow continues, as it would earlier. A response is only returned if it's recieved within 30 seconds.

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+            return data
+
+    open_lock(batch_id)  # initilaze the request
+
+    res_data = process_request(
+        req_body, headers, event, batch_id, write_uri, destination_driver
+    )
+
+    # Write data to s3 or return data synchronously
+    if write_uri:
```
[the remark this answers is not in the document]
sfc-gh-nlele: There weren't any actual modifications to this part actually (this in [main](https://github.com/Snowflake-Labs/geff/blob/e601e97a1aa90649ebc18f15a66cae1a1b8e8a6b/lambda_src/lambda_function.py#L148))

lambda_src/lambda_function.py
```diff
@@ -97,15 +102,115 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
     write_uri = headers.get('write-uri')
+
     LOG.debug(f'sync_flow() received destination: {write_uri}.')
 
     if write_uri:
         destination_driver = import_module(
             f'geff.drivers.destination_{urlparse(write_uri).scheme}'
         )
+
+    while True:
```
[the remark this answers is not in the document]
sfc-gh-nlele: Was trying to implement this and just wanted to confirm: `is_lockable` should just check for a state indicating a lock right?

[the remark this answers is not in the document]
sfc-gh-afedorov: it's a use-once lock for a given batch id which goes from "lockable" to "locked" to "unlocked". in that model `is_lockable` should check for state indicating a lock is in the first state.

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: ah, I see. looks like the diff is confused by the`process_request` addition? would it be possible to move that definition on top of the function in which it is used?

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: Will this leave the lock in a permanently open state?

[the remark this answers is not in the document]
sfc-gh-nlele: Not permanently actually, the TTL of 30m will cause the item to be deleted afterwards, DynamoDB deletes items within 48h of expiration.

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-nlele: Seems like that. Will move it 👍

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: per conversation, I think we agreed on a lock having states "lockable" then "initialized" then "released"

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: per conversation, this is too long -- if we get a 1mb response in 40s, we don't want the user to have to wait 10m to get a "your request has timed out" error

lambda_src/lambda_function.py
```diff
@@ -143,9 +152,65 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+                                ]
+                                for rn, *args in req_body['data']
+                            ]
+                        }
+                    )
+                    gateway_timeout_response = {
+                        'statusCode': 200,
+                        'body': error_dumps,
+                    }
+                    finish_batch_processing(batch_id, gateway_timeout_response)
```
sfc-gh-afedorov, line 206: I think you can replace this with a `return`, as long as the API Gateway times out in 30s, the lambda invocation will be abandoned / retried and it doesn't matter what you respond.

lambda_src/lambda_function.py
```diff
@@ -155,30 +220,80 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+            except ClientError as ce:
+                if ce.response['Error']['Code'] == 'ValidationException':
+                    LOG.error(ce)
+                    error_dumps = dumps(
+                        {
+                            'data': [
+                                [
+                                    rn,
+                                    {
+                                        'error': f'Response size ({len(dumps(response))} bytes) too large to be stored in the backend.',
```
sfc-gh-afedorov, line 239: "response size for batch id ... is too large (... bytes gzip'd). the batch size is {len(...)} and this row's response size is ... bytes"

lambda_src/lambda_function.py
```diff
@@ -155,30 +220,80 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
             'isBase64Encoded': True,
-            'headers': {'Content-Encoding': 'gzip'}
+            'headers': {'Content-Encoding': 'gzip'},
         }
+        end_time = timer()
+        if response and (end_time - start_time) > 20:
+            LOG.debug('Storing the response in lock cache.')
+            try:
+                finish_batch_processing(batch_id, response)  # write the response
+            except ClientError as ce:
```
sfc-gh-afedorov, line 230: move `ClientError` login into `dynamodb.py` and re-raise a generic error

lambda_src/lambda_function.py
```diff
@@ -155,30 +220,80 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+                                [
+                                    rn,
+                                    {
+                                        'error': f'Response size ({len(dumps(response))} bytes) too large to be stored in the backend.',
+                                        'response_hash': md5(
+                                            dumps(response, sort_keys=True).encode()
+                                        ).hexdigest(),
+                                    },
+                                ]
+                                for rn, *args in req_body['data']
```
sfc-gh-afedorov, line 245: maybe `res_data` instead of `req_body['data']` and then you can reconstruct the response sizes?

lambda_src/lambda_function.py
```diff
@@ -209,7 +209,7 @@ def sync_flow(event: Any, context: Any = None) -> Optional[ResponseType]:
         end_time = timer()
         if response and (end_time - start_time) > 20:
```
sfc-gh-pkommini, line 210: Could you put this into a constant with a descriptive name. We should generally try to avoid stray numbers like this as it make the code less readable.

sfc-gh-nlele: Updated it 👍

lambda_src/lambda_function.py
sfc-gh-nlele: Something like below maybe? Would it be fine, or does it seem cluttered?
Screenshot 2023-01-06 at 10 26 39 PM

lambda_src/batch_locking_backends/__init__.py
sfc-gh-afedorov, line 1: docstring would help here. maybe something like —

> these backends helps lock batches so that they are not processed twice when the call driver takes >30s and the API Gateway times out and Snowflake retries the request. each batch has a lock that goes through three states:
> 1. un-initialized batch has not yet been seen by GEFF and so a lock does not exist
> 2. initialized batch is "processing" and has a "locked" lock while the call driver is running
> 3. finished batch means a call driver has responded or timed out and the lock is "unlocked" and stores the response
>
> the modules should expose:
> * `is_batch_initialized(batch_id)` to return `false` for state (1) and `true` otherwise
> * `initialize_batch(batch_id)` to move a lock from (1) to (2)
> * `is_batch_processing(batch_id)` to return `true` for state (2) and `false` otherwise
> * `finish_batch_processing(batch_id, response)` to move lock to state (3) and store a response value
> * `get_response_for_batch(batch_id)` to return the response value for the batch on stage (3) and `None` otherwise
>
> after a batch is finalized, locks should exist for at least 24h to allow for debugging

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+
+AWS_REGION = os.environ.get(
+    "AWS_REGION", "us-west-2"
+)  # Placeholder while in dev TODO: change as variable/header
+DYNAMODB_TABLE = os.environ.get("DYNAMODB_TABLE_NAME")
+DYNAMODB_RESOURCE = boto3.resource("dynamodb", region_name=AWS_REGION)
+if DYNAMODB_TABLE:
+    table = DYNAMODB_RESOURCE.Table(DYNAMODB_TABLE)
+
+TTL = 1800
```
sfc-gh-afedorov, line 18: let's default to 24h and allow for an environment variable to override this

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+import boto3
+from json import dumps
+from hashlib import md5
+
+from botocore.exceptions import ClientError
+from ..utils import LOG, ResponseType
+
+AWS_REGION = os.environ.get(
+    "AWS_REGION", "us-west-2"
+)  # Placeholder while in dev TODO: change as variable/header
```
sfc-gh-afedorov, line 12: not sure I understand this "TODO" -- what will it change to in prod?

lambda_src/lambda_function.py
```diff
@@ -5,11 +5,25 @@
...
+from .batch_locking_backends.dynamodb import (
+    initialize_batch,
+    is_batch_initialized,
+    is_batch_processing,
+    get_response_for_batch,
+    finish_batch_processing,
+)
+
+RESPONSE_SIZE_LIMIT = 6_291_556
+RESPONSE_TIME_BEFORE_LOCK_CACHE_STORAGE = 20
```
sfc-gh-afedorov, line 25: say `SECONDS` instead of `RESPONSE_TIME` so that it's more clear

lambda_src/lambda_function.py
```diff
@@ -5,11 +5,25 @@
...
+from .utils import LOG, create_response, format, invoke_process_lambda, ResponseType
+from .batch_locking_backends.dynamodb import (
+    initialize_batch,
+    is_batch_initialized,
+    is_batch_processing,
+    get_response_for_batch,
+    finish_batch_processing,
+)
+
+RESPONSE_SIZE_LIMIT = 6_291_556
```
sfc-gh-afedorov, line 24: maybe `LAMBDA_RESPONSE_MAX_BYTES`?

lambda_src/lambda_function.py
```diff
@@ -155,30 +207,59 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+    """
+    error_dumps = dumps(
+        {
+            'data': [
+                [
+                    rn,
+                    {
+                        'error': (
+                            f'Response size ({size_exceeded_response_length} bytes) '
+                            'exceeded maximum allowed payload size (6291556 bytes).'
```
sfc-gh-afedorov, line 248: either remove the global variable above bytes count above or use it here

sfc-gh-afedorov (changes requested):
this seems like it should work but there's a couple of readability changes that will help a lot especially as this code evolves. happy to take 30m if I'm missing how any of these are non-trivial.

lambda_src/batch_locking_backends/dynamodb.py
sfc-gh-nlele: I'd replicated it from [here](https://github.com/Snowflake-Labs/geff/blob/09721ec1c09c009f9bab6d3da061fa5fe87f664f/lambda_src/drivers/destination_s3.py#L16), with the assumption that this instance will need the change too.

lambda_src/batch_locking_backends/dynamodb.py
sfc-gh-nlele: Sure 👍

lambda_src/lambda_function.py
```diff
@@ -155,30 +207,59 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
             'statusCode': 200,
             'body': b64encode(compress(data_dumps.encode())).decode(),
             'isBase64Encoded': True,
-            'headers': {'Content-Encoding': 'gzip'}
+            'headers': {'Content-Encoding': 'gzip'},
         }
+        end_time = timer()
+        if (
+            response
```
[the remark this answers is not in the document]
sfc-gh-nlele: True yes

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    try:
+        table.put_item(Item={"batch_id": batch_id, "response": response, "ttl": 1800})
+    except ClientError as ce:
+        if ce.response['Error']['Code'] == 'ValidationException' and res_data:
+            LOG.error(ce)
+            size_exceeded_response = {
+                'statusCode': 200,
+                'body': dumps({
```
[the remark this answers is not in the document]
sfc-gh-nlele: Added

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+def is_batch_processing(batch_id: Text):
+    """
+    Check if a batch id is being processed already, i.e is locked.
+    """
+    return get_response_for_batch(batch_id) == LOCKED
+
+
+def is_batch_initialized(batch_id: Text):
+    """
+    Check if a batch id has been initialized in the batch-locking table.
```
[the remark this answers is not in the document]
sfc-gh-nlele: Sure 👍

lambda_src/lambda_function.py
```diff
@@ -83,38 +97,35 @@ def async_flow_poll(destination: Text, batch_id: Text) -> Dict[Text, Any]:
         return {'statusCode': 202}
 
 
-def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
+def process_batch(
```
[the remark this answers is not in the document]
sfc-gh-nlele: > if I recall my last feedback, it was that we should not pass in all of event and re-get the keys. instead pass in the variables and move the headers pre-processing outside this function

Passed the below as arguments instead, with the headers being already processed in `sync_flow`.

```
processed_headers: Dict[Text, Any],
write_uri: Text,
batch_id: Text,
req_body: Text,
event_path: Text,
```

> this is really hard to read so it might need another read, but it seems like it's only called once? not a super strong preference, but why not just inline this code?

This suggestion came from Prasanth actually.

lambda_src/lambda_function.py
```diff
@@ -83,38 +97,35 @@ def async_flow_poll(destination: Text, batch_id: Text) -> Dict[Text, Any]:
...
+) -> List[List[Any]]:
     """
-    Handles the synchronous part of the generic lambda flows.
+    Processes a request and returns the result data.
 
     Args:
-        event (Any): This the event object as received by the lambda_handler()
-        context (Any): Has the function context. Defaults to None.
+        event (Any): This is the event object as received by the lambda_handler().
+        destination_driver (Optional[ModuleType]): The destination driver such as S3. Defaults to None.
```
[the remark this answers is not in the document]
sfc-gh-nlele: Changed the description.

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+    try:
+        response = item["Item"]["response"]
+    except KeyError:
+        return None
+    return response
+
+
+def is_batch_processing(batch_id: Text):
+    """
+    Check if a batch id is being processed already, i.e is locked.
```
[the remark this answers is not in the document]
sfc-gh-nlele: I thought this would make it clearer what locking is in this context, since we refer to 'locking' in the module name.

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    table.put_item(Item={"batch_id": batch_id, "response": "-1", "ttl": 1800})
+
+
+def get_response_for_batch(batch_id: Text):
+    """
+    Retreive response for a batch id
+    """
+    item = table.get_item(Key={"batch_id": batch_id})
+    try:
```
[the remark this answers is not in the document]
sfc-gh-nlele: Make sense 👍

lambda_src/lambda_function.py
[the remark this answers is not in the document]
sfc-gh-afedorov: > This suggestion came from Prasanth actually.

Do you recall the rationale? Usually the point of sub-procedures as an abstraction is to have them be called more than once.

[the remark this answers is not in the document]
sfc-gh-nlele: It was mostly that the function was less readable and maintainable. I thought it made sense, since the original function is nearly 100 LOC and has a bunch of moving parts.

lambda_src/lambda_function.py
sfc-gh-nlele: Not sure if I understood this one actually, could you elaborate a bit more please?

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,127 @@
...
+
+AWS_REGION = os.environ.get(
+    'AWS_REGION', 'us-west-2'
+)  # Placeholder while in dev TODO: change as variable/header
+DYNAMODB_TABLE = os.environ.get('DYNAMODB_TABLE_NAME')
+TTL = os.environ.get('DYNAMODB_TABLE_TTL', 86400)
+
+DYNAMODB_RESOURCE = boto3.resource('dynamodb', region_name=AWS_REGION)
+if DYNAMODB_TABLE:
+    table = DYNAMODB_RESOURCE.Table(DYNAMODB_TABLE)
```
sfc-gh-afedorov, line 33: how will the rest of the code work if table is not set?

lambda_src/batch_locking_backends/dynamodb.py
```diff
@@ -0,0 +1,127 @@
...
+from botocore.exceptions import ClientError
+from ..utils import LOG, ResponseType
+
+AWS_REGION = os.environ.get(
+    'AWS_REGION', 'us-west-2'
+)  # Placeholder while in dev TODO: change as variable/header
+DYNAMODB_TABLE = os.environ.get('DYNAMODB_TABLE_NAME')
+TTL = os.environ.get('DYNAMODB_TABLE_TTL', 86400)
+
+DYNAMODB_RESOURCE = boto3.resource('dynamodb', region_name=AWS_REGION)
```
sfc-gh-afedorov, line 31: you're only using this in one place -- would it make sense to inline it?

sfc-gh-nlele: Makes sense, inlined it.

lambda_src/lambda_function.py
```diff
@@ -146,9 +147,63 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+    write_uri = headers.get('write-uri')
+    destination_driver = (
+        import_module(f'geff.drivers.destination_{urlparse(write_uri).scheme}')
+        if write_uri
+        else None
+    )
+
+    LOG.debug(f'sync_flow() received destination: {write_uri}.')
+
+    if BATCH_LOCKING_ENABLED and not destination_driver:
```
sfc-gh-nlele, line 180: @sfc-gh-afedorov Would it be better to pass a generic `BATCH_LOCKING_ENABLED` boolean instead of a module specific `DYNAMODB_TABLE`?

lambda_src/lambda_function.py
```diff
@@ -86,41 +96,32 @@ def async_flow_poll(destination: Text, batch_id: Text) -> Dict[Text, Any]:
...
-
-    LOG.debug(f'sync_flow() received destination: {write_uri}.')
-
-    if write_uri:
-        destination_driver = import_module(
-            f'geff.drivers.destination_{urlparse(write_uri).scheme}'
-        )
     res_data = []
 
     for row_number, *args in req_body['data']:
```
sfc-gh-nlele, line 120: pass req_body['data']

lambda_src/lambda_function.py
```diff
@@ -146,9 +147,63 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        if not is_batch_initialized(batch_id):
+            initialize_batch(batch_id)
+        else:
+            while is_batch_processing(batch_id):
+                if (timer() - start_time) > SECONDS_BEFORE_GATEWAY_TIMEOUT:
+                    return None
+
+            return get_response_for_batch(batch_id)
+
+    driver_kwargs = {
```
sfc-gh-nlele, line 190: Add type sig

lambda_src/lambda_function.py
sfc-gh-afedorov: Yes, and mention it in the interface if it's something all modules should export

sfc-gh-afedorov (dismissed):

lambda_src/lambda_function.py
```diff
@@ -146,9 +147,64 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        event (Any): This the event object as received by the lambda_handler()
+        context (Any): Has the function context. Defaults to None.
+
+    Returns:
+        Optional[ResponseType]: Represents the response status and data.
+    """
+    LOG.debug('Destination header not found in a POST and hence using sync_flow().')
+    headers = event['headers']
+    req_body = loads(event['body'])
+    req_body_data: List[List[Union[int, str]]] = req_body['data']
```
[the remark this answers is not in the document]
sfc-gh-nlele: I think I'd assigned the type `int`/`str` considering we expect things like headers to be a string too (like below) while processing the arguments. 

```
[[0, 'https://67f3-34-214-158-144.ngrok.io', 'GET', '{"Content-Type": "application/json"}']]
```

But true, `array`/`variant` would still be valid inputs. Will keep the inner type to be `any`.

PR #60, opened 2022-10-11
Add json body as an auth option

sfc-gh-amathew:
Added some changes to the https driver to allow json to be passed for authentication to an api. The code changes have been tested with the amathew-geff-lambda function that is using the geff image with the tag '0.2.0-test'

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -101,7 +101,26 @@ def process_row(
         req_data: Optional[bytes] = (
             json if json.startswith('{') else dumps(parse_header_dict(json))
         ).encode()
+     elif json.startswith('arn'):
+        json if json.startswith('arn') 
+        auth = decrypt_if_encrypted(json)
+        req_auth = (
+        loads(auth)
+        if auth and auth.startswith('{')
+        else parse_header_dict(auth)
+        if auth
+        else {} 
+        ) 
+        auth_host = req_auth.get('host') 
+
+        # We reject the request if the 'auth' is present but doesn't match the pinned host.
+        if auth_host and req_host and auth_host != req_host:
+            raise ValueError(
+                "Requests can only be made to host provided in the auth header."
+            )
         req_headers['Content-Type'] = 'application/json'
+        if('ClientId' and 'secret' in req_auth):
+            req_data={"clientId": str(req_auth['ClientId']),"secret": str(req_auth['secret'])}
     else:
         req_data = None if data is None else data.encode()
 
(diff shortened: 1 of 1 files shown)


lambda_src/drivers/process_https.py
```diff
@@ -98,10 +98,23 @@ def process_row(
     req_method: str = method.upper()
 
     if json:
+        
         req_data: Optional[bytes] = (
             json if json.startswith('{') else dumps(parse_header_dict(json))
         ).encode()
         req_headers['Content-Type'] = 'application/json'
+        if json.startswith('arn'):
```
sfc-gh-afedorov, line 106: this check already exists in `decrypt_if_encrypted` -- we want it there since there's no need for this code to worry about the details of how secrets are stored.

lambda_src/drivers/process_https.py
```diff
@@ -98,10 +98,23 @@ def process_row(
     req_method: str = method.upper()
 
     if json:
+
```
sfc-gh-afedorov, line 101: don't add vestigial changes, esp this one since it breaks formatting guidelines from PEP-8. if it helps, install a formatter that will enforce pep8 like https://black.readthedocs.io/en/stable/

lambda_src/drivers/process_https.py
```diff
@@ -98,10 +98,23 @@ def process_row(
...
+        
         req_data: Optional[bytes] = (
             json if json.startswith('{') else dumps(parse_header_dict(json))
         ).encode()
         req_headers['Content-Type'] = 'application/json'
+        if json.startswith('arn'):
+            data=(decrypt_if_encrypted(json))
+            data=loads(data)
+            json_host=data['host']
+            if json_host and req_host and json_host != req_host:
```
sfc-gh-afedorov, line 110: I think it's better to include `json-body` as an auth option in the part that already has this check, could you add it there and re-use that structure rather than including a second one?

lambda_src/drivers/process_https.py
```diff
@@ -98,10 +98,23 @@ def process_row(
...
+                raise ValueError(
+                "Requests can only be made to host provided in the auth header."
+                )
+            if('clientId' and 'secret' in data):
```
sfc-gh-afedorov, line 114: these are details of a single EF, so they shouldn't go into a GEFF driver, since those should be Generic

sfc-gh-afedorov (changes requested):
a few comments but also please keep PR's as draft until the automated feedback is resolved

please include a descriptive title and description but no need to mention the specifics of what partner you are integrating with

PR #61, opened 2022-10-13
Add json-body option to auth header

sfc-gh-amathew:
Certain API require credentials to be passed part of the payload as opposed to headers that GEFF currently supports. This PR is to allow users to provide credentials to GEFF via payload. 

This change has been tested using an API that requires a payload in the body to authenticate.

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -28,7 +28,6 @@ def process_row(
     method: Text = 'get',
     headers: Text = '',
     kwargs: Union[Dict, Text] = '',
-    #auth to auth header
     auth: Text = None,
     params: Text = '',
     verbose: bool = False,
@@ -59,6 +58,7 @@ def process_row(
 
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
+    auth_body = {}
 
     # We look for an auth header and if found, we parse it from its encoded format
     if auth:
@@ -93,23 +93,25 @@ def process_row(
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
         elif 'body' in req_auth:
-            pre_data = req_auth['body']
-            req_data = pre_data.encode('utf-8')
+            auth_body = req_auth['body']
         
     # query, nextpage_path, results_path
     req_params: str = params
     req_results_path: str = results_path
     req_cursor: str = cursor
     req_method: str = method.upper()
-
+        
+    if auth_body or json:
+        req_headers['Content-Type'] = 'application/json'
     if json:
         req_data= (
-            json if json.startswith('{') else dumps(parse_header_dict(json))
+            json if json.startswith('{') else dumps(parse_header_dict(json.update(auth_body)))
         ).encode()
-        req_headers['Content-Type'] = 'application/json'
+    elif auth_body:
+        req_data = auth_body.encode('utf-8')        
     else:
         req_data = None if data is None else data.encode()
-
+    
     if req_params:
         req_url += f'?{req_params}'
 
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -28,7 +28,9 @@ def process_row(
     method: Text = 'get',
     headers: Text = '',
     kwargs: Union[Dict, Text] = '',
-    auth: Text = None,
+    #auth to auth header
+    auth_header: Text = None,
+    auth_body :Optional[Text] = None,
     params: Text = '',
     verbose: bool = False,
     cursor: Text = '',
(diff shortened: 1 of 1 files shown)


lambda_src/drivers/process_https.py
```diff
@@ -90,15 +92,18 @@ def process_row(
             req_headers['authorization'] = req_auth['authorization']
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
-
+        elif 'body' in req_auth:
+            pre_data=req_auth['body']
+            req_data1=pre_data.encode('utf-8')
```
sfc-gh-pkommini, line 97: What is `req_data1`? Consider a more descriptive name?

lambda_src/drivers/process_https.py
```diff
@@ -90,15 +92,18 @@ def process_row(
             req_headers['authorization'] = req_auth['authorization']
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
-
+        elif 'body' in req_auth:
+            pre_data=req_auth['body']
```
sfc-gh-pkommini, line 96: variables should always have space between them and operators. So change to:

```
pre_data = req_auth['body']
req_data = pre_data.encode('utf-8')
```

lambda_src/drivers/process_https.py
sfc-gh-amathew: It was actually req_data I put req_data1 by mistake.

lambda_src/drivers/process_https.py
sfc-gh-amathew: Done

lambda_src/drivers/process_https.py
```diff
@@ -54,6 +55,7 @@ def process_row(
             else {}
         ).items()
     }
+    req_data: Optional[bytes]=None
```
sfc-gh-pkommini, line 58: Formatting please. you need to leave spaces before variables and operators and values.

lambda_src/drivers/process_https.py
```diff
@@ -54,9 +54,11 @@ def process_row(
...
```
sfc-gh-afedorov, line 57: this change is no longer necessary

lambda_src/drivers/process_https.py
```diff
@@ -54,9 +54,11 @@ def process_row(
...
 
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
+    auth_body: Optional[str] = None
```
sfc-gh-afedorov, line 61: this change is no longer necessary

lambda_src/drivers/process_https.py
```diff
@@ -90,21 +92,31 @@ def process_row(
...
+        elif 'body' in req_auth:
+            json_body = (
+                loads(req_auth['body'])
+                if isinstance(req_auth['body'], str)
+                else req_auth['body']
+            )    
+            if json:
+                raise ValueError(f"both auth body and json present")
+            else:
+                json = json_body
```
sfc-gh-afedorov, line 104: on second read, we only use this variable once, so it's probably best to in-line the above code here

lambda_src/drivers/process_https.py
```diff
@@ -90,21 +92,31 @@ def process_row(
...
+        
     # query, nextpage_path, results_path
     req_params: str = params
     req_results_path: str = results_path
     req_cursor: str = cursor
     req_method: str = method.upper()
-
+        
     if json:
-        req_data: Optional[bytes] = (
```
sfc-gh-afedorov, line 101: this change is now vestigial

lambda_src/drivers/process_https.py
```diff
@@ -90,21 +92,31 @@ def process_row(
...
+        req_data= (
             json if json.startswith('{') else dumps(parse_header_dict(json))
-        ).encode()
-        req_headers['Content-Type'] = 'application/json'
```
sfc-gh-afedorov, line 104: this seems like a mistaken change that introduces a syntax error

sfc-gh-afedorov (changes requested):
looks like there's a lot of stray changes here, some harmless but others less so -- let's have a quick meet to go over workflow (I think we're recommending VSCode, altho Sublime is good, too) and install the plugins and maybe GitHub Desktop to help you construct more idiomatic diffs for review.

lambda_src/drivers/process_https.py
sfc-gh-amathew: The change is removed.

lambda_src/drivers/process_https.py
sfc-gh-amathew: The changes is removed.

sfc-gh-afedorov (approved):
lgtm but please provide testing done to description before merging

lambda_src/drivers/process_https.py
```diff
@@ -90,6 +90,15 @@ def process_row(
             req_headers['authorization'] = req_auth['authorization']
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
+        elif 'body' in req_auth:   
+            if json:
+                raise ValueError(f"auth 'body' key and json param are both present")
```
sfc-gh-pkommini, line 95: Is it the case that auth body and json should not simultaneously? This seems to be a contract with the GEFF clients and I think should be documented in the wiki along with being present in code.

sfc-gh-amathew: Documented this in the wiki.

sfc-gh-pkommini (approved):
lgtm! I think we should document the new structure that `auth` header can take and additionally assumptions around it not being present with `json`. Not a blocking change but might be good to do before deploying.

PR #63, opened 2022-12-13
Fix lambda response length check on dict

sfc-gh-nlele:
This PR fixes the response to be returned when the API response size exceeds Lambda's limits. 

Testing: 

Constructed a response of ~11 MB after compression, the output in Snowflake:

Screenshot 2023-01-06 at 6 00 43 PM

### DIFF
--- a/lambda_src/lambda_function.py
+++ b/lambda_src/lambda_function.py
@@ -159,22 +159,25 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
         }
 
     if len(dumps(response)) > 6_000_000:
-        response = dumps(
-            {
-                'data': [
-                    [
-                        rn,
-                        {
-                            'error': (
-                                f'Response size ({len(dumps(response))} bytes) will likely'
-                                'exceeded maximum allowed payload size (6291556 bytes).'
-                            )
-                        },
+        response = {
+            'statusCode': 202,
+            'body': dumps(
+                {
+                    'data': [
+                        [
+                            rn,
+                            {
+                                'error': (
+                                    f'Response size ({len(dumps(response))} bytes) will likely '
+                                    'exceeded maximum allowed payload size (6291556 bytes).'
+                                )
+                            },
+                        ]
+                        for rn, *args in req_body['data']
                     ]
-                    for rn, *args in req_body['data']
-                ]
-            }
-        )
+                }
+            ),
+        }
     return response
 
 
--- a/lambda_src/lambda_function.py
+++ b/lambda_src/lambda_function.py
@@ -155,10 +155,10 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
             'statusCode': 200,
             'body': b64encode(compress(data_dumps.encode())).decode(),
             'isBase64Encoded': True,
-            'headers': {'Content-Encoding': 'gzip'}
+            'headers': {'Content-Encoding': 'gzip'},
         }
 
-    if len(response) > 6_000_000:
+    if len(dumps(response)) > 6_000_000:
         response = dumps(
             {
                 'data': [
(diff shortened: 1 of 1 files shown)


lambda_src/lambda_function.py
```diff
@@ -155,26 +155,29 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
-                        rn,
-                        {
-                            'error': (
-                                f'Response size ({len(response)} bytes) will likely'
-                                'exceeded maximum allowed payload size (6291556 bytes).'
-                            )
-                        },
+    if len(dumps(response)) > 6_000_000:
+        response = {
+            'statusCode': 202,
```
sfc-gh-afedorov, line 163: why `statusCode` 202? I think from perspective of EF invocation, this means the response is complete, so it should be 200.

sfc-gh-afedorov (changes requested):
minor change but lgtm and ty

add a description of how (if?) you tested that this works

update title to be more specific and descriptive of context, e.g. start with "fix" to show that you're fixing a bug, and describe which response you are checking (lambda response)

lambda_src/lambda_function.py
sfc-gh-nlele: Makes sense 👍 Will change it.

sfc-gh-nlele:
> minor change but lgtm and ty
> 
> add a description of how (if?) you tested that this works
> 
> update title to be more specific and descriptive of context, e.g. start with "fix" to show that you're fixing a bug, and describe which response you are checking (lambda response)

Would it be better to stick the req locking branch for this, since this is a minor change and it will be more coherent there?

lambda_src/lambda_function.py
```diff
@@ -155,26 +155,29 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
-                'data': [
-                    [
-                        rn,
-                        {
-                            'error': (
-                                f'Response size ({len(response)} bytes) will likely'
-                                'exceeded maximum allowed payload size (6291556 bytes).'
-                            )
-                        },
+    if len(dumps(response)) > 6_000_000:
```
sfc-gh-afedorov, line 161: per [this blog post](https://zaccharles.medium.com/deep-dive-lambdas-response-payload-size-limit-8aedba9530ed), I think we can change the constant to 6_291_556

lambda_src/lambda_function.py
```diff
@@ -155,26 +155,29 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
...
+        response = {
+            'statusCode': 200,
+            'body': dumps(
+                {
+                    'data': [
+                        [
+                            rn,
+                            {
+                                'error': (
+                                    f'Response size ({len(dumps(response))} bytes) will likely '
```
sfc-gh-afedorov, line 171: so as to not do `len(dumps(...))` twice, let's put it in a variable `response_length`

with the exact constant, we can remove the word "likely" here

sfc-gh-afedorov (approved):
lgtm but a few minor touches

lambda_src/lambda_function.py
sfc-gh-nlele: Makes sense 👍

lambda_src/lambda_function.py
sfc-gh-nlele: > with the exact constant, we can remove the word "likely" here

Removed it  👍 

> so as to not do len(dumps(...)) twice, let's put it in a variable response_length

This actually turns into a function in [#59](https://github.com/Snowflake-Labs/geff/pull/59/files#diff-2b9d704e4645a913e70521e3fa194921702cf9e9aee4ce6573403adcd80c2d58R225) so it might have to be repeated I suppose?

sfc-gh-afedorov: In #59 you can pass the integer length instead of the string, as well.

sfc-gh-nlele: Oo right, made the change here and tested it again, okay to merge?

sfc-gh-afedorov (approved):
yup, lgtm

PR #64, opened 2023-01-10
Add support for storing non-json responses, and a hashsum flag in paths

sfc-gh-nlele:
This PR adds support for the storage of non-json/file responses to S3 and storing their sha256 hashsum in the file paths, thorugh a flag.

The hashum flag can be added like: `s3://geff-bucket/foo/{sha2}/bar`, which then gets replaced by the hash of the response: `s3://geff-bucket/foo/d5e2f10fe2e5d7ed2001c5513c22c56330519342e76f6467919310497d5c5ce6/bar`

Testing:

Example flow for driver storage:

Getting file names and links:

Screenshot 2023-02-15 at 12 47 50 PM

External function:
```
CREATE OR REPLACE SECURE EXTERNAL FUNCTION TEST_FUNCTION(url string, destination string)
RETURNS VARIANT
RETURNS NULL ON NULL INPUT
VOLATILE
API_INTEGRATION= API_INTEGRATION
HEADERS=(
   'url'     = '{0}'
   'headers' = '{"Content-Type": "application/json", "Accept": "application/json"}'
   'method'  = 'GET'
   'destination-uri' = 's3://geff-rlm-dev-dev-bucket/driver-hashsums/{1}/{sha2}/{1}'
)
AS 'https://a1b2c3d4.execute-api.us-west-2.amazonaws.com/dev/https';
```
Call:
```
SELECT test_function(driver_link, driver_name) from driver_destinations;
```
Storage in S3:

Screenshot 2023-02-15 at 1 05 23 PM

Driver w/ hash in the path:

Screenshot 2023-02-15 at 1 09 04 PM

Response in Snowflake:

Screenshot 2023-02-17 at 3 44 08 PM

### DIFF
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -5,6 +5,7 @@ from typing import Any, AnyStr, Dict, Generator, List, Optional, Text, Tuple, Un
 from urllib.parse import urlparse
 from time import strftime
 import re
+from hashlib import md5, sha3_256
 
 import boto3
 from botocore.exceptions import ClientError
@@ -31,10 +32,7 @@ def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
     """
     LOG.debug(f'destination from header is {destination}')
     parsed_url = urlparse(destination)
-    return (
-        parsed_url.netloc,
-        strftime(parsed_url.path[1:])  # remove leading slash
-    )
+    return (parsed_url.netloc, strftime(parsed_url.path[1:]))  # remove leading slash
 
 
 def estimated_record_size(records: List[Dict[Text, Any]]) -> float:
@@ -116,6 +114,8 @@ def write(
             prefixed_filename,
             encoded_datum,
         ),
+        'md5_hash': md5(encoded_datum.encode('utf-8')).hexdigest(),
+        'sha3_hash': sha3_256(encoded_datum.encode('utf-8')).hexdigest(),
         'uri': s3_uri,
     }
 
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -105,7 +105,7 @@ def write(
         prefixed_filename = prefix.format(hash=encoded_datum_hash)
     else:
         encoded_datum = (
-            '\n'.join(json.dumps(d) for d in datum)
+            '\n'.join(json.dumps(d) for d in datum)  # type: ignore
             if isinstance(datum, list)
             else json.dumps(datum, default=str)
         )
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -9,7 +9,7 @@ from hashlib import sha256
 
 import boto3
 from botocore.exceptions import ClientError
-from ..utils import LOG, hash_format
+from ..utils import LOG, lazy_format
 
 SAMPLE_SIZE: int = 10
 MAX_JSON_FILE_SIZE: int = 15 * 1024 * 1024 * 1024
@@ -87,7 +87,7 @@ def initialize(destination: Text, batch_id: Text):
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
     prefix_folder = re.sub(r'/[^/]*$', '/', prefix) if '/' in prefix else ''
-    prefix_folder = re.sub(r'/{\s*sha2\s*}.*', '/', prefix_folder, flags=re.IGNORECASE)
+    prefix_folder = re.sub(r'/{\s*hash\s*}.*', '/', prefix_folder, flags=re.IGNORECASE)
 
     if prefix_folder:
         write_to_s3(bucket, prefix_folder, content)
@@ -102,8 +102,8 @@ def write(
     bucket, prefix = parse_destination_uri(destination)
     if isinstance(datum, bytes):
         encoded_datum = datum
-        prefixed_filename = hash_format(
-            prefix, encoded_datum, sha2=lambda x: sha256(x).hexdigest()
+        prefixed_filename = lazy_format(
+            prefix, hash=lambda: sha256(encoded_datum).hexdigest()
         )
     else:
         encoded_datum = (
@@ -115,7 +115,7 @@ def write(
 
     s3_uri = f's3://{bucket}/{prefixed_filename}'
 
-    return {
+    response = {
         'response': write_to_s3(
             bucket,
             prefixed_filename,
@@ -123,6 +123,9 @@ def write(
         ),
         'uri': s3_uri,
     }
+    if isinstance(encoded_datum, bytes):
+        response['sha256_hash'] = sha256(encoded_datum).hexdigest()
+    return response
 
 
 def finalize(
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -104,7 +104,6 @@ def write(
         prefixed_filename = hash_format(
             prefix, encoded_datum, sha3_256=lambda x: sha3_256(x).hexdigest()
         )
-        LOG.debug(f"hash: {sha3_256(encoded_datum).hexdigest()}")
     else:
         encoded_datum = (
             '\n'.join(json.dumps(d) for d in datum)
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -89,7 +89,7 @@ def initialize(destination: Text, batch_id: Text):
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
     prefix_folder = re.sub(r'/[^/]*$', '/', prefix) if '/' in prefix else ''
     prefix_folder = (
-        re.sub(r'/{\s*sha2\s*}/$', '/', prefix_folder, flags=re.IGNORECASE)
+        re.sub(r'/{\s*sha2\s*}.*', '/', prefix_folder, flags=re.IGNORECASE)
         if HASH_PARAM in prefix_folder.lower()
         else prefix_folder
     )
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -5,7 +5,7 @@ from typing import Any, AnyStr, Dict, Generator, List, Optional, Text, Tuple, Un
 from urllib.parse import urlparse
 from time import strftime
 import re
-from hashlib import md5, sha3_256
+from hashlib import sha256
 
 import boto3
 from botocore.exceptions import ClientError
@@ -19,6 +19,7 @@ AWS_REGION = os.environ[
 S3_CLIENT = boto3.client('s3', region_name=AWS_REGION)
 MANIFEST_FILENAME = 'MANIFEST.json'
 MANIESTS_FOLDER_NAME = 'meta'
+HASH_PARAM = '{sha2}'
 
 
 def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
@@ -87,6 +88,11 @@ def initialize(destination: Text, batch_id: Text):
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
     prefix_folder = re.sub(r'/[^/]*$', '/', prefix) if '/' in prefix else ''
+    prefix_folder = (
+        re.sub(r'/{\s*sha2\s*}/$', '/', prefix, flags=re.IGNORECASE)
+        if HASH_PARAM in prefix_folder.lower()
+        else prefix_folder
+    )
 
     if prefix_folder:
         write_to_s3(bucket, prefix_folder, content)
@@ -102,7 +108,7 @@ def write(
     if isinstance(datum, bytes):
         encoded_datum = datum
         prefixed_filename = hash_format(
-            prefix, encoded_datum, sha3_256=lambda x: sha3_256(x).hexdigest()
+            prefix, encoded_datum, sha2=lambda x: sha256(x).hexdigest()
         )
     else:
         encoded_datum = (
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -19,6 +19,7 @@ AWS_REGION = os.environ[
 S3_CLIENT = boto3.client('s3', region_name=AWS_REGION)
 MANIFEST_FILENAME = 'MANIFEST.json'
 MANIESTS_FOLDER_NAME = 'meta'
+HASH_PARAM = 'hash'
 
 
 def parse_destination_uri(destination: Text) -> Tuple[Text, Text]:
@@ -87,8 +88,12 @@ def initialize(destination: Text, batch_id: Text):
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
     prefix_folder = re.sub(r'/[^/]*$', '/', prefix) if '/' in prefix else ''
-    prefix_folder = re.sub(r'/{\s*hash\s*}.*', '/', prefix_folder, flags=re.IGNORECASE)
-
+    prefix_folder = re.sub(
+        r'/{\s*' + HASH_PARAM + r'\s*}.*',
+        '/',
+        prefix_folder,
+        flags=re.IGNORECASE,
+    )
     if prefix_folder:
         write_to_s3(bucket, prefix_folder, content)
 
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -107,6 +107,7 @@ def write(
         else json.dumps(datum, default=str).encode()
     )
     encoded_datum_hash = sha256(encoded_datum).hexdigest()
+
     prefixed_filename = (
         f'{prefix}{batch_id}_row_{row_index}.data.json'
         if prefix.endswith('/')
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -1,7 +1,7 @@
 import json
 import os
 from random import sample
-from typing import Any, Dict, Generator, List, Optional, Text, Tuple, Union, AnyStr
+from typing import Any, AnyStr, Dict, Generator, List, Optional, Text, Tuple, Union
 from urllib.parse import urlparse
 from time import strftime
 import re
--- a/lambda_src/drivers/destination_s3.py
+++ b/lambda_src/drivers/destination_s3.py
@@ -1,7 +1,7 @@
 import json
 import os
 from random import sample
-from typing import Any, AnyStr, Dict, Generator, List, Optional, Text, Tuple, Union
+from typing import Any, Dict, Generator, List, Optional, Text, Tuple, Union
 from urllib.parse import urlparse
 from time import strftime
 import re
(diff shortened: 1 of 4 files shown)


lambda_src/drivers/process_https.py
```diff
@@ -146,7 +148,10 @@ def process_row(
                 if 'Date' in response_headers
                 else None
             )
-            response_body = loads(raw_response)
+            try:
```
sfc-gh-afedorov, line 151: check if response_headers['Content-Type'] == 'application/json', instead

lambda_src/drivers/destination_s3.py
```diff
@@ -101,13 +100,19 @@ def write(
...
+        prefixed_filename = hash_format(
+            prefix, encoded_datum, sha2=lambda x: sha256(x).hexdigest()
+        )
+    else:
+        encoded_datum = (
+            '\n'.join(json.dumps(d) for d in datum)
+            if isinstance(datum, list)
+            else json.dumps(datum, default=str)
+        )
+        prefixed_filename = f'{prefix}{batch_id}_row_{row_index}.data.json'
```
sfc-gh-nlele, line 114: We'd want to maintain uniformity by hash-formatting in either of the above cases right?

lambda_src/drivers/destination_s3.py
```diff
@@ -89,6 +87,7 @@ def initialize(destination: Text, batch_id: Text):
     # Regex captures characters after and including the rightmost '/' in a path,
     # which are then replaced with a '/', e.g. '/a/b/c' -> '/a/b/'
     prefix_folder = re.sub(r'/[^/]*$', '/', prefix) if '/' in prefix else ''
+    prefix_folder = re.sub(r'/{\s*sha2\s*}.*', '/', prefix_folder, flags=re.IGNORECASE)
```
sfc-gh-afedorov, line 90: this seems a little bit too specific -- could you change it to something that doesn't explicitly say "sha2" in it? if you'd prefer to keep it, let's make it a global variable so that it's defined once and used twice making it impossible for the two to be out of sync.

sfc-gh-afedorov: maybe `r'/{[^}]}.*`'"?

lambda_src/drivers/destination_s3.py
sfc-gh-afedorov: it would be nice to have, but not necessary for this use-case, so can also wait

lambda_src/drivers/destination_s3.py
```diff
@@ -101,13 +100,19 @@ def write(
...
     bucket, prefix = parse_destination_uri(destination)
-    encoded_datum = (
-        '\n'.join(json.dumps(d) for d in datum)
-        if isinstance(datum, list)
-        else json.dumps(datum, default=str)
-    )
+    if isinstance(datum, bytes):
+        encoded_datum = datum
+        prefixed_filename = hash_format(
+            prefix, encoded_datum, sha2=lambda x: sha256(x).hexdigest()
```
sfc-gh-afedorov, line 106: instead of taking it explicitly, when defining a lambda you can "enclose" variables in the scope to create a [closue](https://www.geeksforgeeks.org/python-closures/), i.e. you can do `lambda: sha256(encoded_datum).hexdigest()` here and it will do what you'd expect

lambda_src/drivers/process_https.py
```diff
@@ -17,7 +18,7 @@ def make_basic_header(auth):
 
 
 def parse_header_dict(value):
-    return {k: decrypt_if_encrypted(v) for k, v in parse_qsl(value)}
+    return {k: v for k, v in parse_qsl(value)}
```
sfc-gh-afedorov, line 21: rebase on master?

sfc-gh-afedorov (changes requested):
didn't finish reading all of it, but largely lgtm so far, commented on a few changes and could you rebase on master to separate out the changes that were merged earlier?

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Makes sense 👍

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: > this seems a little bit too specific -- could you change it to something that doesn't explicitly say "sha2" in it? if you'd prefer to keep it, let's make it a global variable so that it's defined once and used twice making it impossible for the two to be out of sync.

Will do, I'd kept the name similar to Snowflake's [SHA](https://docs.snowflake.com/en/sql-reference/functions/sha2#syntax) function. Maybe something more generic such as `{hash}`?

lambda_src/drivers/process_https.py
sfc-gh-nlele: Had already rebased this branch actually, the PR just seems confused like you said, a new PR on this branch doesn't contain this change.

lambda_src/utils.py
```diff
@@ -107,3 +107,29 @@ def invoke_process_lambda(event: Any, lambda_name: Text) -> Dict[Text, Any]:
...
+    def lazy_eval(match_obj):
+        key = match_obj.group(1)
+        if key in kwargs:
+            return str(kwargs[key](datum))
+        return match_obj.group(0)
+
+    return re.sub(r'\{(\w+)\}', lazy_eval, fmt_str)
+
+
+def process_row_params_format(headers, args):
```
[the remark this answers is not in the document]
sfc-gh-nlele: Ah right, makes sense.

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: > maybe `r'/{[^}]}.*'`?

Did you mean `r'/{[^}].*`?

I suppose there wouldn't be cases where a `{foobar}` in the path is intended to be anything besides a parameter, so should be fine?

sfc-gh-nlele: Had pushed a [global variable](https://github.com/Snowflake-Labs/geff/pull/64/commits/0c0773309bc50ed6ede340e481f8a181fe509a6f) for the hash param since we don't support any other brace'd paramaters, but keeping it more generic seems to make more sense now.

lambda_src/utils.py
```diff
@@ -107,3 +107,29 @@ def invoke_process_lambda(event: Any, lambda_name: Text) -> Dict[Text, Any]:
 
     # Returns 202 on success if InvocationType = 'Event'
     return lambda_response
+
+
+def hash_format(fmt_str, datum, **kwargs):
+    '''
+    Format filepath with the hashsum of a datum
```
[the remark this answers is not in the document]
sfc-gh-nlele: Makes sense 👍

lambda_src/drivers/destination_s3.py
```diff
@@ -101,23 +99,31 @@ def write(
...
     bucket, prefix = parse_destination_uri(destination)
-    encoded_datum = (
-        '\n'.join(json.dumps(d) for d in datum)
-        if isinstance(datum, list)
-        else json.dumps(datum, default=str)
-    )
+    if isinstance(datum, bytes):
+        encoded_datum = datum
+        encoded_datum_hash = sha256(encoded_datum).hexdigest()
+        prefixed_filename = lazy_format(prefix, hash=lambda: encoded_datum_hash)
```
sfc-gh-nlele, line 105: Used `encoded_datum_hash` in the lambda to not re-calculate the hash on line [125](https://github.com/Snowflake-Labs/geff/pull/64/files#diff-361e1d3c9eb36aa3cc3dfcb657f34db3c71d7b22a6f5b3d10fb2a37a3bfd81faR125)

lambda_src/drivers/process_https.py
```diff
@@ -32,6 +33,7 @@ def process_row(
     verbose: bool = False,
     cursor: Text = '',
     results_path: Text = '',
+    write_uri: Optional[Text] = '',
```
sfc-gh-afedorov, line 36: why this?

lambda_src/lambda_function.py
```diff
@@ -19,6 +24,27 @@
 DESTINATION_URI_HEADER = 'sf-custom-destination-uri'
 
 
+def process_row_params_format(
+    headers: Dict[Text, Text], args: Tuple[Any]
+) -> Dict[Text, Any]:
+    '''
+    Format parameters in headers with arguments from a row.
```
sfc-gh-afedorov, line 31: a little ambivalent about this. it doesn't explain why `-` is turned to `_` and `sf-custom` is removed. maybe add that it's in preparation to be used as parameters to the call driver, or call it `process_headers_with_row`?

lambda_src/lambda_function.py
```diff
@@ -33,19 +59,22 @@ def async_flow_init(event: Any, context: Any) -> Dict[Text, Any]:
...
     LOG.debug(f'async_flow_init() received destination: {destination}.')
 
     destination_driver = import_module(
         f'geff.drivers.destination_{urlparse(destination).scheme}'
     )
-    # Ignoring style due to dynamic import
-    destination_driver.initialize(destination, batch_id)  # type: ignore
+    # Ignoring style due to dynamic imports
+    for rn, *args in req_body['data']:
+        destination = process_row_params_format(headers, args)['destination_uri']
```
sfc-gh-afedorov, line 74: `destination_uri` is guaranteed to be there because it's the transformed version of the value in `DESTINATION_URI_HEADER`, and everything else is ignored? that seems like a bit of an indirect coupling. why not just call `format(destination, args)`?

then you can also just inline `process_row_params_format` abstraction, which seems like a tough one to name and explain concisely, anyway.

lambda_src/lambda_function.py
```diff
@@ -97,9 +126,9 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
     LOG.debug('Destination header not found in a POST and hence using sync_flow().')
     headers = event['headers']
     req_body = loads(event['body'])
-
+    write_uri = headers.get('sf-custom-write-uri')
```
sfc-gh-afedorov, line 129: I think this might have intentionally been set to something the user could not pass, but I might be mis-remembering. why the change?

sfc-gh-afedorov (commented):
a quick pass, can review more thoroughly again tomorrow

lambda_src/drivers/destination_s3.py
```diff
@@ -101,23 +99,31 @@ def write(
...
+    response = {
         'response': write_to_s3(
             bucket,
             prefixed_filename,
             encoded_datum,
         ),
         'uri': s3_uri,
     }
+    if isinstance(encoded_datum, bytes):
+        response['sha256_hash'] = encoded_datum_hash
```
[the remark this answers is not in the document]
sfc-gh-nlele: True true :)

lambda_src/drivers/destination_s3.py
```diff
@@ -101,23 +99,31 @@ def write(
     row_index: int,
 ) -> Dict[Text, Any]:
     bucket, prefix = parse_destination_uri(destination)
-    encoded_datum = (
-        '\n'.join(json.dumps(d) for d in datum)
-        if isinstance(datum, list)
-        else json.dumps(datum, default=str)
-    )
+    if isinstance(datum, bytes):
```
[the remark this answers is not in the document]
sfc-gh-nlele: Ah right, updated 👍

lambda_src/drivers/process_https.py
sfc-gh-nlele: This was for below, where the params get unpacked and `write_uri` is still present:  
https://github.com/Snowflake-Labs/geff/blob/d74930640267968553989195d663ca020af6955c/lambda_src/lambda_function.py#L154

But will remove `write_uri` from the params dictionary before it gets unpacked now, since it seems like the better option.

lambda_src/lambda_function.py
sfc-gh-nlele: Will do 👍

lambda_src/lambda_function.py
sfc-gh-nlele: > destination_uri is guaranteed to be there because it's the transformed version of the value in DESTINATION_URI_HEADER, and everything else is ignored? that seems like a bit of an indirect coupling. why not just call format(destination, args)?

Makes sense 👍

lambda_src/lambda_function.py
sfc-gh-nlele: This was so that we could format the `write-uri`, as the destionation header is popped earlier and `process_row_params_format` filters on `sf-custom`

sfc-gh-nlele: Removed the `sf-custom`  from `write-uri` and used a `format` like [here](https://github.com/Snowflake-Labs/geff/pull/64#discussion_r1116158870). 

https://github.com/Snowflake-Labs/geff/blob/7da2fef21098b2ed3a522aeea92ac8fb20d0c6f9/lambda_src/lambda_function.py#L108-L126

lambda_src/lambda_function.py
```diff
@@ -33,19 +38,22 @@ def async_flow_init(event: Any, context: Any) -> Dict[Text, Any]:
     LOG.debug('Found a destination header and hence using async_flow_init().')
 
     headers = event['headers']
+    req_body = loads(event['body'])
     batch_id = headers[BATCH_ID_HEADER]
     destination = headers[DESTINATION_URI_HEADER]
-    headers.pop(DESTINATION_URI_HEADER)
```
sfc-gh-afedorov, line 38: I don't think this breaks anything, but it also doesn't seem to do anything helpful. if you want to refactor this, let's maybe do something like 

```python
destination = headers['write-uri'] = headers.pop(DESTINATION_URI_HEADER)
```

as that simplifies the flow a little bit?

lambda_src/lambda_function.py
```diff
@@ -33,19 +38,22 @@ def async_flow_init(event: Any, context: Any) -> Dict[Text, Any]:
...
 
     destination_driver = import_module(
         f'geff.drivers.destination_{urlparse(destination).scheme}'
     )
-    # Ignoring style due to dynamic import
-    destination_driver.initialize(destination, batch_id)  # type: ignore
+    # Ignoring style due to dynamic imports
+    for rn, *args in req_body['data']:
+        parameterized_destination = format(destination, args)
+        destination_driver.initialize(parameterized_destination, batch_id)  # type: ignore
```
sfc-gh-afedorov, line 54: I think you realizes this changes the contract between the main GEFF lambda processor and the write drivers, calling initialize once per row instead of once per EF call. I think we talked about this so I imagine you have a reason here, but I'm not sure what it is, and it seems it could add a lot of overhead at scale.

might it make sense to start out by requiring that the initialize call not be parametrized by the row args, only by the keywords we're adding, and avoid the change to the driver interface?

sfc-gh-afedorov (commented):
some minor comments, I think we're close to done!

lambda_src/drivers/process_https.py
```diff
@@ -141,7 +142,12 @@ def process_row(
                 if 'Date' in response_headers
                 else None
             )
-            response_body = loads(raw_response)
+
+            if response_headers['Content-Type'] == 'application/json':
+                response_body = loads(raw_response)
```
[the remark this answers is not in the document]
sfc-gh-nlele: True true 👍

lambda_src/drivers/destination_s3.py
```diff
@@ -89,35 +87,43 @@ def initialize(destination: Text, batch_id: Text):
...
+        encoded_datum = datum
+        encoded_datum_hash = sha256(encoded_datum).hexdigest()
+        prefixed_filename = prefix.format(hash=encoded_datum_hash)
+    else:
+        encoded_datum = (
+            '\n'.join(json.dumps(d) for d in datum)  # type: ignore
```
sfc-gh-nlele, line 108: Since `encoded_datum` can be of type `str` or `bytes`, mypy produces the following error:

```
error: Incompatible types in assignment (expression has type "str", variable has type "bytes")
```

If the type for `encoded_datum` is set to `Union[bytes, str, AnyStr]`, mypy produces this error:

```
error: Value of type variable "AnyStr" of "write_to_s3" cannot be "Union[bytes, str]"
```

This is seems to be an open bug: https://github.com/python/mypy/issues/1533

So best to just ignore the type for now? @sfc-gh-afedorov

sfc-gh-nlele:
Retested the changes to make sure that nothing breaks.

lambda_src/lambda_function.py
```diff
@@ -128,10 +132,10 @@ def sync_flow(event: Any, context: Any = None) -> Dict[Text, Any]:
             row_result = process_row(*path, **process_row_params)
             LOG.debug(f'Got row_result for URL: {process_row_params.get("url")}.')
 
-            if write_uri:
+            if parameterized_write_uri:
                 # Write s3 data and return confirmation
                 row_result = destination_driver.write(  # type: ignore
-                    write_uri, batch_id, row_result, row_number
+                    parameterized_write_uri, batch_id, row_result, row_number
```
sfc-gh-afedorov, line 138: seems it would be easier to just pass `format(write_uri)` here instead of creating a separate variable for it?

sfc-gh-afedorov (changes requested):
nice. all these changes seem like they'll do what they intend. a few stylistic changes below and one feature change that came to mind. the interpretation of the destination parameter in the write driver really shouldn't change based on the header of the call driver response. described a good way to change that below, but lmk if it seems off?

realize there's a lot of iterations here, thanks for your patience and lmk if you'd like me to make these changes and re-test.

lambda_src/lambda_function.py
sfc-gh-afedorov: ty, a comment might do well to clarify, altho it does seem relatively straightforward now that it's just a format. there's a couple of reasons I've seen extra variables used, but not seeing them here.

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: > Long story short, If we're going to use mypy it's best not ignore its errors and add an explicit type declaration at the top of the function, like encoded_datum: Union[bytes, str].

True, I'd tried forward declaring the type actually, but it ended up with the AnyStr error. 

> In this case, tho, I think it's best cast the right hand value on line 107 to bytes so that it's aligned with what the boto3 function is documented to receive, computes the hash in both cases, and removes the need for the if conditional and response variable below

Oo that works 👍

sfc-gh-nlele:
> the interpretation of the destination parameter in the write driver really shouldn't change based on the header of the call driver response. described a good way to change that below, but lmk if it seems off?

Ah no, it makese sense

> realize there's a lot of iterations here, thanks for your patience and lmk if you'd like me to make these changes and re-test.

Npnp! Will add the changes 👍

lambda_src/drivers/destination_s3.py
```diff
@@ -89,25 +87,37 @@ def initialize(destination: Text, batch_id: Text):
...
-    datum: Union[Dict, List],
+    datum: Union[Dict, List, bytes],
     row_index: int,
 ) -> Dict[Text, Any]:
     bucket, prefix = parse_destination_uri(destination)
     encoded_datum = (
-        '\n'.join(json.dumps(d) for d in datum)
+        datum
+        if isinstance(datum, bytes)
+        else ('\n'.join(json.dumps(d) for d in datum)).encode()
```
sfc-gh-nlele, line 105: bytes() needs an encoding like `utf-8`, when used with a string, so encoded both of the strings instead

sfc-gh-afedorov: that's probably better, to avoid calling bytes on bytes.

the `.encode()` defaults to `utf-8`. an encoding is necessary because `bytes` are a specific abstraction while `str` is a generic with many encodings possible in memory/on disk/over network.

sfc-gh-afedorov (approved):
lgtm, ty for the patience

lambda_src/drivers/destination_s3.py
sfc-gh-nlele: Yess

sfc-gh-nlele:
Ty! Merging the PR, had re-tested the changes.

PR #65, opened 2023-01-20
Remove kwargs header

sfc-gh-nlele:
This PR removes the `decrypt_if_encrypted` call in the `parse_header_dict` function of the `process_https` driver and the vestigial kwargs header now that we have changed the strategy to always specify how auth secrets should be sent.

Testing:
Retested the changes.

Some examples:
After the change:

1) When using kwargs:

EF:
```
CREATE OR REPLACE SECURE EXTERNAL FUNCTION TEST_FUNCTION(GAMMA STRING, DELTA STRING)
RETURNS VARIANT
RETURNS NULL ON NULL INPUT
VOLATILE
MAX_BATCH_ROWS = 1
API_INTEGRATION=API_INTEGRATION
HEADERS=(
   'url'     = 'https://httpbin.org/get'
   'headers' = '{ "Content-Type": "application/json", "Accept": "application/json"}'
   'method'  = 'GET'
   'kwargs' =  'x={0}&y={1}'
   'headers' = '{ "Gamma": "{x}", "Delta" : "{y}"}'
)
AS 'https://abcde12345.execute-api.us-west-2.amazonaws.com/dev/https';

SELECT TEST_FUNCTION('foo','bar');
```
Result:
Screenshot 2023-01-20 at 6 26 01 PM

2) When using plain headers and no kwargs:

EF:
```
CREATE OR REPLACE SECURE EXTERNAL FUNCTION TEST_FUNCTION(GAMMA STRING, DELTA STRING)
RETURNS VARIANT
RETURNS NULL ON NULL INPUT
VOLATILE
MAX_BATCH_ROWS = 1
API_INTEGRATION=API_INTEGRATION
HEADERS=(
   'url'     = 'https://httpbin.org/get'
   'headers' = '{ "Content-Type": "application/json", "Accept": "application/json"}'
   'method'  = 'GET'
   'headers' = '{ "Gamma": "{0}", "Delta" : "{1}"}'
)
AS 'https://abcde12345.execute-api.us-west-2.amazonaws.com/dev/https';

SELECT TEST_FUNCTION('foo','bar');
```
Result:
Screenshot 2023-01-20 at 6 32 50 PM

3) Using headers for auth:

```
CREATE OR REPLACE SECURE EXTERNAL FUNCTION TEST_DB.DATA.TEST_FUNCTION()
RETURNS VARIANT
RETURNS NULL ON NULL INPUT
VOLATILE
MAX_BATCH_ROWS = 1
API_INTEGRATION=API_INTEGRATION
HEADERS=(
   'url'     = 'https://httpbin.org/headers'
   'headers' = '{ "Content-Type": "application/json", "Accept": "application/json"}'
   'method'  = 'GET'
   'headers' = 'foobar=arn:aws:secretsmanager:us-west-2:123456:secret:test/secret-a1b2c3'
)
AS 'https://abcde12345.execute-api.us-west-2.amazonaws.com/dev/https';

SELECT TEST_FUNCTION();

```

Result:
Screenshot 2023-01-25 at 10 31 25 PM

Before the change:
1) When using kwargs

EF:
```
CREATE OR REPLACE SECURE EXTERNAL FUNCTION TEST_FUNCTION(GAMMA STRING, DELTA STRING)
RETURNS VARIANT
RETURNS NULL ON NULL INPUT
VOLATILE
MAX_BATCH_ROWS = 1
API_INTEGRATION=API_INTEGRATION
HEADERS=(
   'url'     = 'https://httpbin.org/get'
   'headers' = '{ "Content-Type": "application/json", "Accept": "application/json"}'
   'method'  = 'GET'
   'kwargs' =  'x={0}&y={1}'
   'headers' = '{ "Gamma": "{x}", "Delta" : "{y}"}'
)
AS 'https://abcde12345.execute-api.us-west-2.amazonaws.com/dev/https';

SELECT TEST_FUNCTION('foo','bar');
```
Result:
Screenshot 2023-01-20 at 6 42 24 PM

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -27,7 +27,6 @@ def process_row(
     json: Optional[Text] = None,
     method: Text = 'get',
     headers: Text = '',
-    kwargs: Union[Dict, Text] = '',
     auth: Text = None,
     params: Text = '',
     verbose: bool = False,
@@ -43,17 +42,13 @@ def process_row(
         raise ValueError('URL scheme must be HTTPS.')
 
     req_host = u.hostname
-    req_kwargs = parse_header_dict(kwargs)
-    req_headers = {
-        k: v.format(**req_kwargs)
-        for k, v in (
-            loads(headers)
-            if headers.startswith('{')
-            else parse_header_dict(headers)
-            if headers
-            else {}
-        ).items()
-    }
+    req_headers = (
+        loads(headers)
+        if headers.startswith('{')
+        else parse_header_dict(headers)
+        if headers
+        else {}
+    )
 
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
@@ -90,7 +85,7 @@ def process_row(
             req_headers['authorization'] = req_auth['authorization']
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
-        elif 'body' in req_auth:   
+        elif 'body' in req_auth:
             if json:
                 raise ValueError(f"auth 'body' key and json param are both present")
             else:
@@ -204,7 +199,9 @@ def process_row(
 
             next_url = (
                 cursor_value
-                if cursor_value and isinstance(cursor_value, str) and cursor_value.startswith('https://')
+                if cursor_value
+                and isinstance(cursor_value, str)
+                and cursor_value.startswith('https://')
                 else f'{req_url}&{cursor_param}={cursor_value}'
                 if cursor_value
                 else None
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -17,7 +17,7 @@ def make_basic_header(auth):
 
 
 def parse_header_dict(value):
-    return {k: decrypt_if_encrypted(v) for k, v in parse_qsl(value)}
+    return {k: v for k, v in parse_qsl(value)}
 
 
 def process_row(


lambda_src/drivers/process_https.py
```diff
@@ -43,17 +42,13 @@ def process_row(
...
-            if headers.startswith('{')
-            else parse_header_dict(headers)
-            if headers
-            else {}
-        ).items()
-    }
+    req_headers = (
+        loads(headers)
+        if headers.startswith('{')
+        else parse_header_dict(headers)
```
sfc-gh-afedorov, line 48: err, I think as long as we call `parse_header_dict`, we will be decrypting things without checking what they are. could we also remove the `decrypt_if_encrypted` call from there?

there's a second (properly gated) call on line 58, which should be the only one left.

sfc-gh-afedorov (changes requested):

lambda_src/drivers/process_https.py
sfc-gh-nlele: oo right, makes sense. removed it.

lambda_src/drivers/process_https.py
```diff
@@ -204,7 +199,9 @@ def process_row(
 
             next_url = (
                 cursor_value
-                if cursor_value and isinstance(cursor_value, str) and cursor_value.startswith('https://')
+                if cursor_value
```
sfc-gh-afedorov, line 202: ok for now but let's try to move towards a process where these sorts of stray changes don't happen in a commit. when making changes, it's not just about what the code looks like, but what the `git blame` looks like, as understanding the history of the code is a valuable component of how the code is.

sfc-gh-afedorov (approved):
updated description a little bit. lgtm. please re-test and let's keep an eye on Sentry as we deploy this since it's a breaking change for any pipelines using kwargs.

lambda_src/drivers/process_https.py
sfc-gh-nlele: makes sense, will keep this in mind 👍

sfc-gh-pkommini:
We don't pass an argument called `kwargs` from the external function. We just pass key value pairs that end up in `kwargs` which is a dictionary of all optional arguments.

While that changes nothing in the code. I think there is a slight misunderstanding regarding how arguments can be passed into the driver.

sfc-gh-afedorov:
We can't anymore as a result of this change, but we used to pass in kwargs in an earlier iteration of secrets management, e.g. in the Tenable ingestion [here](https://github.com/snowflakedb/seceng-snowhouse-infra/blob/21978f074dbb0025fe7249cddfdebab6d405a84c/api_ingest/tenable.sql#L34). It violated the security assurances we want to make about secrets being stored along with limits on their use that the lambda code respects.

PR #68, opened 2023-04-12
check response content type

sfc-gh-amathew:
This is a small change in the https driver to check the content type of the api response. The earlier check was a direct comparison but the content type from certain api could return the char encoding also. This means that the response would become a byte string when it was supposed to be a dict. 

This can cause errors downstream if external function expect a dict.


sfc-gh-nlele:
Maybe just the `'application/json' in req_headers['Content-Type']` change in this PR?

sfc-gh-afedorov (approved):

PR #70, opened 2023-04-20
Support for Jinja templating

sfc-gh-amathew:
This PR is to add support for rendering authentication strings using Jinja templates.


sfc-gh-afedorov (approved):
tested against some isolated sample code, so I think this should work. we should deploy it to a dev environment before merging, though, just to be doubly sure.

sfc-gh-amathew:
Will test it on dev before merging.

PR #71, opened 2023-05-17
additonal condition for url payload

sfc-gh-amathew:
This PR is to push an additional condition for the data payload for the URL. The change is to check if the value in json is a string in addition to the existing condition to check if the string starts with '{'

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -143,6 +143,7 @@ def process_row(
         req_data: Optional[bytes] = (
             json if json.startswith('{') else dumps(parse_header_dict(json))
         ).encode()
+        req_headers['Content-Type'] = 'application/json'
     else:
         req_data = None if data is None else data.encode()
 
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -9,8 +9,9 @@ from re import match
 from time import time
 from typing import Any, Dict, List, Optional, Text, Union
 from urllib.error import HTTPError, URLError
-from urllib.parse import parse_qsl, urlparse
+from urllib.parse import parse_qsl, urlparse, urlencode
 from urllib.request import Request, urlopen
+
 
 from ..utils import LOG, parse_header_links, pick
 from ..vault import decrypt_if_encrypted
@@ -136,12 +137,12 @@ def process_row(
     req_results_path: str = results_path
     req_cursor: str = cursor
     req_method: str = method.upper()
-
     if json:
         req_data: Optional[bytes] = (
-            json if json.startswith('{') else dumps(parse_header_dict(json))
+            json
+            if (isinstance(json, str) or json.startswith('{}'))
+            else dumps(parse_header_dict(json))
         ).encode()
-        req_headers['Content-Type'] = 'application/json'
     else:
         req_data = None if data is None else data.encode()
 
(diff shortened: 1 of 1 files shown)


sfc-gh-afedorov:
What is this looking to fix?

lambda_src/drivers/process_https.py
```diff
@@ -136,12 +137,12 @@ def process_row(
     req_results_path: str = results_path
     req_cursor: str = cursor
     req_method: str = method.upper()
-
     if json:
         req_data: Optional[bytes] = (
-            json if json.startswith('{') else dumps(parse_header_dict(json))
+            json
+            if (isinstance(json, str) or json.startswith('{}'))
```
sfc-gh-afedorov, line 143: what will this statement do if type of json is a dict?

sfc-gh-afedorov (changes requested):

lambda_src/drivers/process_https.py
sfc-gh-amathew: The else condition will convert a dict to a JSON string if the json variable has a dict.

lambda_src/drivers/process_https.py
```diff
@@ -136,12 +137,12 @@ def process_row(
     req_results_path: str = results_path
     req_cursor: str = cursor
     req_method: str = method.upper()
-
     if json:
         req_data: Optional[bytes] = (
-            json if json.startswith('{') else dumps(parse_header_dict(json))
+            json
+            if (isinstance(json, str) or json.startswith('{'))
```
sfc-gh-afedorov, line 143: I think you mean to say `and` not `or` here. also please use pyblack or manually remove the extra parentheses

lambda_src/drivers/process_https.py
```diff
@@ -136,12 +137,12 @@ def process_row(
...
+            else dumps(parse_header_dict(json))
         ).encode()
-        req_headers['Content-Type'] = 'application/json'
```
sfc-gh-afedorov, line 144: why remove this?

sfc-gh-afedorov:
still confused — in the function you are changing, `json` is defined as `Optional[Text]` so inside of the `if json` statement it should be `Text`, and reading the code it seems like it's assuming it's of the format (A) `a=1&b=2` or (B) `{"a": 1, "b": 2}` and it normalizes those to JSON bytes.

the `isinstance(json, str)` seems like it will always evaluate to true in this context, which will break the case of parameters being passed in as format (A).

it's also not clear to me what it is that you are looking to fix

lambda_src/drivers/process_https.py
sfc-gh-amathew: We are passing content-type in the external function.

lambda_src/drivers/process_https.py
```diff
@@ -123,10 +124,10 @@ def process_row(
         elif 'headers' in req_auth:
             req_headers.update(req_auth['headers'])
         elif 'body' in req_auth:
-            if json:
-                raise ValueError(f"auth 'body' key and json param are both present")
+            if data:
```
sfc-gh-afedorov, line 127: keep the json error so that a Snowflake user can't override what's in the secret by passing a `json` header, either

lambda_src/drivers/process_https.py
sfc-gh-afedorov: we could, but this behavior was intentional -- it mimics the way Python's requests lib works, and it might be something our EF's currently rely on, which could break them. it's also not related to this bug fix, so should be in a separate review if you want to migrate away from it.

sfc-gh-afedorov (approved):
lgtm ty

sfc-gh-afedorov (approved):
lgtm ty

PR #77, opened 2023-12-13
Fix HMAC auth to work with pagination cursors

sfc-gh-afedorov:
move auth code to execute inside of the pagination loop

add unit test to make sure the headers are set independently for each page

### DIFF
--- a/lambda_src/drivers/process_https.py
+++ b/lambda_src/drivers/process_https.py
@@ -83,76 +83,13 @@ def process_row(
     req_headers.setdefault('User-Agent', 'GEFF 1.0')
     req_headers.setdefault('Accept-Encoding', 'gzip')
 
-    # We look for an auth header and if found, we parse it from its encoded format
-    if auth:
-        auth = render_jinja_template(
-            decrypt_if_encrypted(auth),
-            {
-                'path': parsed_url.path,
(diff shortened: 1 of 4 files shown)


lambda_src/drivers/process_https.py
```diff
@@ -71,130 +83,136 @@ def process_row(
...
 
     LOG.debug('Starting pagination.')
     while next_url:
-        LOG.debug(f'next_url is {next_url}.')
-        req = Request(next_url, method=req_method, headers=req_headers, data=req_data)
+        parsed_url = urlparse(next_url)
+
+        # We look for an auth header and if found, we parse it from its encoded format
+        if auth:
+            auth = render_jinja_template(
```
sfc-gh-afedorov, line 109: to debug this started by looking through the free variables in the expression and blocks to abstract those relationships as a function.

in this specific expression,`(auth, parsed_url, method, ...)` is used as input and then overwrites `auth`. this is fine if it's run once, but inside of a loop, we need to preserve the original `auth` that holds the template. it's often questionable practice to re-use a variable name like this and this is a decent example of why.

sfc-gh-nlele (approved):

sfc-gh-nlele:
lgtm!

PR #82, opened 2026-08-19
Sandbox HTTPS driver Jinja env & harden HTTPS authentication

sfc-gh-afedorov:
## Summary

This PR captures a security review of the HTTPS authentication path performed via Codex/gpt-5.6-Sol.

- sandbox Jinja auth-template rendering and expose only explicitly registered helpers
- prevent credentials from following cross-origin pagination links and redirects
- sign auth templates with the HTTP method that is sent to server

## Jinja template sandboxing

The `auth` value can be supplied directly or loaded from AWS Secrets Manager, then is evaluated as a Jinja template before request host validation. A standard `Environment` permits traversal from Jinja globals into Python function globals, which can turn control of an auth template into code execution in the Lambda runtime.

The renderer now uses `SandboxedEnvironment`, clears Jinja's default globals, and adds back only the helpers intentionally registered by GEFF. Regression tests cover unsafe traversal through a registered helper and removal of default globals.

Context: https://huggingface.co/blog/agent-intrusion-technical-timeline#vector-2-jinja2-template-injection

## Cross-origin credential handling

Pagination links and automatic redirects could previously carry authentication headers, custom headers, or an auth-generated request body to another origin. Python's `urllib` redirect handling also preserves request headers by default.

Requests now compare scheme, hostname, and effective port with the initial request origin. Cross-origin pagination retains only GEFF's default transport headers, skips auth-template application, and does not reuse an auth-generated body. A custom redirect handler applies the same policy to automatic redirects while preserving credentials for same-origin redirects.

## Request signing consistency

Auth templates previously received the caller-provided method string while the outgoing request used its uppercase form. Templates now receive the normalized method that is actually sent, preventing signatures over `get` when the wire request uses `GET`.

## Testing

- Python 3.9: `python -m pytest tests` — 15 passed
- `mypy==0.930 --config-file .mypy.ini lambda_src/` — no issues


sfc-gh-afedorov:
@codex we good?

sfc-gh-bkou (approved):

